PR #53, opened 2026-06-27
feat(desktop): native desktop app launcher

### LINKED ISSUE #40: [idea] package into native app
`uv tool install orionbelt_ontology_builder` collects all deps fine only it doesn't find any executables that can be 'installed', so just drops the installation procedure there.

Because the app is based on `streamlit` it shouldn't be much of a problem, apparently `streamlit-desktop-app` and `stlite` are the 2 standard ways to wrap streamlits into a native app.

Obviously it would be a lot more convenient to run, as there is no need to start/stop the server as well as close/open the browser tab, and it would really give the app the finishing touch it deserves (desktop integration as a bonus :) ).

ralfbecher:
## What

Adds an opt-in **native desktop window** for the app, addressing the remaining native desktop request in #40 (the `uv tool install` console launcher already shipped in v1.6.2).

Uses [`streamlit-desktop-app`](https://pypi.org/project/streamlit-desktop-app/) (pywebview + a real Streamlit server), the dependable route from the issue discussion. Because it runs a real Streamlit server, every dependency works as-is, including `streamlit-local-storage` autosave (the part that would have been risky under stlite/Pyodide).

```bash
pip install "orionbelt-ontology-builder[desktop]"
orionbelt-ontology-builder-desktop    # opens a native window
```

## Changes

- New `orionbelt_ontology_builder/desktop.py`: `run()` opens the app in a native window (1280x800), reusing the in-package `streamlit_entry.py` already used by the console launcher. If the optional extra is missing it prints a friendly install hint and exits 1.
- New optional `desktop` extra (`streamlit-desktop-app>=0.3.3`) and a second console script `orionbelt-ontology-builder-desktop`.
- New `tests/test_desktop.py`: entry-point registration, the launch path (mocked), and the missing-dependency fallback.
- README: "Run as a native desktop app" subsection.
- Version bump 1.7.2 to 1.8.0.

## Notes

- Fully opt-in. The plain install and the existing `orionbelt-ontology-builder` (browser) command are unchanged.
- The extra installs and runs on Python 3.10 to 3.13 (`streamlit-desktop-app` metadata is `>=3.9`, no upper bound).
- Verified locally: full suite (262 passed), ruff clean, and the native window opens and autosaves on Python 3.13 / macOS.

## Scope

This covers the native window only. Standalone per-OS installers (PyInstaller + CI artifacts) are tracked as a follow-up in #54, so this intentionally does not close #40.

Refs #40

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 <p align="center"><strong>A browser-based ontology workbench built with Streamlit and rdflib</strong></p>
 
 [![GitHub stars](https://img.shields.io/github/stars/ralforion/orionbelt-ontology-builder?style=social)](https://github.com/ralforion/orionbelt-ontology-builder)
-[![Version 1.7.2](https://img.shields.io/badge/version-1.7.2-purple.svg)](https://github.com/ralforion/orionbelt-ontology-builder/releases)
+[![Version 1.8.0](https://img.shields.io/badge/version-1.8.0-purple.svg)](https://github.com/ralforion/orionbelt-ontology-builder/releases)
 [![Python 3.10+](https://img.shields.io/badge/python-3.10+-blue.svg)](https://www.python.org/downloads/)
 [![License: BSL 1.1](https://img.shields.io/badge/License-BSL_1.1-orange.svg)](https://github.com/ralforion/orionbelt-ontology-builder/blob/main/LICENSE)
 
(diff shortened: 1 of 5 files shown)


SuperCowProducts:
Hello @ralfbecher, I was able to get `orionbelt-ontology-builder-desktop` working only after `uv pip install pyside6` inside the virtual environment, could you add it do the deps included in `orionbelt-ontology-builder[desktop]` package?

SuperCowProducts:
Just wanted to show system integration is very good once a desktop entry is created to the executable in the virtual environment with `favicon.png` (included in the repo) added:
Screenshot_20260627_163630

Issue #74, opened 2026-06-29
[packaging] 1.9.3 is the same as 1.9.2


SuperCowProducts:
Looks like the `v1.9.3` changes didn't make it to the [PyPI package](https://pypi.org/project/orionbelt-ontology-builder/1.9.3/) for some reason!

<details>

<summary>
Logs
</summary>

```bash
yalihupokn@zoqu-endiman:~/Repositories/OB/orionbelt-ontology-builder$ cd ~/Downloads/
yalihupokn@zoqu-endiman:~/Downloads$ python -m venv .venv
yalihupokn@zoqu-endiman:~/Downloads$ . .venv/bin/activate
(.venv) yalihupokn@zoqu-endiman:~/Downloads$ pip install orionbelt-ontology-builder[qt]
Collecting orionbelt-ontology-builder[qt]
  Downloading orionbelt_ontology_builder-1.9.3-py3-none-any.whl.metadata (17 kB)
WARNING: orionbelt-ontology-builder 1.9.3 does not provide the extra 'qt'
Collecting streamlit==1.49.1 (from orionbelt-ontology-builder[qt])
  Downloading streamlit-1.49.1-py3-none-any.whl.metadata (9.5 kB)
Collecting rdflib>=7.0.0 (from orionbelt-ontology-builder[qt])
  Downloading rdflib-7.6.0-py3-none-any.whl.metadata (12 kB)
Collecting owlrl>=6.0.2 (from orionbelt-ontology-builder[qt])
  Downloading owlrl-7.1.4-py3-none-any.whl.metadata (3.8 kB)
Collecting networkx>=3.0 (from orionbelt-ontology-builder[qt])
  Using cached networkx-3.6.1-py3-none-any.whl.metadata (6.8 kB)
Collecting pyvis>=0.3.2 (from orionbelt-ontology-builder[qt])
  Downloading pyvis-0.3.2-py3-none-any.whl.metadata (1.7 kB)
Collecting streamlit-local-storage>=0.0.25 (from orionbelt-ontology-builder[qt])
  Downloading streamlit_local_storage-0.0.25-py3-none-any.whl.metadata (3.8 kB)
Collecting altair!=5.4.0,!=5.4.1,<6,>=4.0 (from streamlit==1.49.1->orionbelt-ontology-builder[qt])
  Downloading altair-5.5.0-py3-none-any.whl.metadata (11 kB)
[...]

Issue #68, opened 2026-06-28
[deps] `qtpy`, `pyside6`


SuperCowProducts:
`orionbelt-ontology-builder-desktop` fails otherwise

SuperCowProducts:
ah never mind, in `v1.9.0` it doesn't seem to work even after installing `pyside6`. 

<details>

<summary>
Log
</summary>

```bash
(orionbelt-ontology-builder) yalihupokn@zoqu-endiman:~/Repositories/OB/orionbelt-ontology-builder$ orionbelt-ontology-builder-desktop 

  You can now view your Streamlit app in your browser.

  URL: http://localhost:45223

[pywebview] QT cannot be loaded
Traceback (most recent call last):
  File "/home/yalihupokn/.local/share/uv/tools/orionbelt-ontology-builder/lib/python3.13/site-packages/webview/guilib.py", line 49, in import_qt
    import webview.platforms.qt as guilib
  File "/home/yalihupokn/.local/share/uv/tools/orionbelt-ontology-builder/lib/python3.13/site-packages/webview/platforms/qt.py", line 23, in <module>
    from qtpy import QtCore
ModuleNotFoundError: No module named 'qtpy'
ERROR:pywebview:QT cannot be loaded
Traceback (most recent call last):
  File "/home/yalihupokn/.local/share/uv/tools/orionbelt-ontology-builder/lib/python3.13/site-packages/webview/guilib.py", line 49, in import_qt
    import webview.platforms.qt as guilib
  File "/home/yalihupokn/.local/share/uv/tools/orionbelt-ontology-builder/lib/python3.13/site-packages/webview/platforms/qt.py", line 23, in <module>
    from qtpy import QtCore
ModuleNotFoundError: No module named 'qtpy'
[pywebview] GTK cannot be loaded
Traceback (most recent call last):
  File "/home/yalihupokn/.local/share/uv/tools/orionbelt-ontology-builder/lib/python3.13/site-packages/webview/guilib.py", line 37, in import_gtk
    import webview.platforms.gtk as guilib
  File "/home/yalihupokn/.local/share/uv/tools/orionbelt-ontology-builder/lib/python3.13/site-packages/webview/platforms/gtk.py", line 20, in <module>
    import gi
ModuleNotFoundError: No module named 'gi'
ERROR:pywebview:GTK cannot be loaded
Traceback (most recent call last):
[...]

PR #75, opened 2026-06-29
feat(desktop): add qt/gtk backend extras and deterministic backend selection

### LINKED ISSUE #73: [enhancement] Optional [qt] / [gtk] desktop backend extras
Follow-up to #68 (suggested by @SuperCowProducts).

## Background

The desktop app renders through pywebview, which needs a GUI backend. Today the `desktop` extra hardcodes the Qt backend (`pywebview[pyside6]`, which pulls qtpy + PySide6). pywebview also supports a GTK backend (`pywebview[gtk]`, via PyGObject/`gi` + WebKitGTK).

## Proposal

Offer the backend as a choice via separate extras, keeping `desktop` as the working default:

```toml
desktop = [   # unchanged default (Qt)
    "streamlit-desktop-app>=0.3.3",
    "pywebview[pyside6]; sys_platform != 'darwin'",
]
qt = [        # explicit Qt, same as desktop
    "streamlit-desktop-app>=0.3.3",
    "pywebview[pyside6]; sys_platform != 'darwin'",
]
gtk = [       # GTK backend instead of Qt
    "streamlit-desktop-app>=0.3.3",
    "pywebview[gtk]; sys_platform != 'darwin'",
]
```

Users who prefer GTK (or want to avoid the large Qt/Chromium download) could then `pip install "orionbelt-ontology-builder[gtk]"`.

## Caveats / open questions

- `pywebview[gtk]` needs system packages pip cannot install (gobject-introspection, libgirepository, webkit2gtk), so installs fail more often than Qt unless the distro already provides them. Docs would need to spell this out.
- Three overlapping extras add maintenance and documentation surface.
- CI ideally tests both backends; GTK is awkward to verify headless, so coverage may stay Qt-only.

Priority: nice-to-have. The current Qt `desktop` extra already works for the common case.

ralfbecher:
Closes #73

## What

- Two new install extras: **`qt`** (explicit alias of `desktop`, PySide6) and **`gtk`** (Linux-only, PyGObject).
- The launcher now **selects the pywebview backend deterministically** instead of relying on import order.

## Why

pywebview picks its backend by import order, not by what you installed. On Linux it tries **GTK before Qt** regardless, so:
- Qt-only users saw a scary `[pywebview] GTK cannot be loaded` traceback on every launch before it fell back to Qt (visible in #68's log).
- Simply adding a `gtk` extra would make the active backend depend on import order, not on the extra you chose.

## How

`_preferred_gui()` checks importability (no heavy imports) and exports `PYWEBVIEW_GUI`:
- `gi` present -> `gtk`; else `qtpy` present -> `qt`
- respects a user-set `PYWEBVIEW_GUI`; returns `None` on macOS (Cocoa stays default)

`pyproject.toml`:
```toml
qt  = [ "streamlit-desktop-app>=0.3.3", "pywebview[pyside6]; sys_platform != 'darwin'" ]
gtk = [ "streamlit-desktop-app>=0.3.3", "pywebview[gtk]; sys_platform == 'linux'" ]
```
We keep `pywebview[pyside6]` (LGPL PySide6) for `qt`, not pywebview's `qt` extra (which would pull a conflicting PyQt6).

## Notes / caveats

- `pywebview[gtk]` needs system packages pip cannot install (gobject-introspection, libgirepository, the WebKit typelib). README documents the `apt` line.
- Tests cover backend selection (gtk/qt/none/override/macOS) and that `run()` exports `PYWEBVIEW_GUI`. Full suite: 313 passed.
- No GTK CI job: a resolution-only check adds little over docs, and GTK headless is the most flake-prone piece. The existing `desktop-extra` (Qt) job still guards the default.

Bumps version to 1.10.0 (new user-facing capability).

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 <p align="center"><strong>A browser-based ontology workbench built with Streamlit and rdflib</strong></p>
 
 [![GitHub stars](https://img.shields.io/github/stars/ralforion/orionbelt-ontology-builder?style=social)](https://github.com/ralforion/orionbelt-ontology-builder)
-[![Version 1.9.3](https://img.shields.io/badge/version-1.9.3-purple.svg)](https://github.com/ralforion/orionbelt-ontology-builder/releases)
+[![Version 1.10.0](https://img.shields.io/badge/version-1.10.0-purple.svg)](https://github.com/ralforion/orionbelt-ontology-builder/releases)
 [![Python 3.10+](https://img.shields.io/badge/python-3.10+-blue.svg)](https://www.python.org/downloads/)
 [![License: BSL 1.1](https://img.shields.io/badge/License-BSL_1.1-orange.svg)](https://github.com/ralforion/orionbelt-ontology-builder/blob/main/LICENSE)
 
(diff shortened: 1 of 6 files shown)


ralfbecher:
Also folds in the fix for #74: bumps `APP_VERSION` (which was left at 1.9.2 in the published 1.9.3 because the edit was never staged) and adds a CI guard asserting `APP_VERSION` and the README badge stay in sync with `pyproject.toml`. This PR closes #74 as well.

PR #84, opened 2026-06-30
fix(viz): preserve zoom/pan on click; cache layout on re-mount

ralfbecher:
Fixes the graph resetting its zoom/pan when you click a node, plus the flicker when toggling the panel.

## The bug

Clicking a node triggers a Streamlit rerun, which re-rendered the component and **rebuilt the whole vis-network from scratch — resetting the user's zoom/pan**. Toggling the panel additionally re-mounted the component and re-ran the physics layout (flicker/jump).

## The fix

1. **Don't rebuild on selection.** `renderGraph` now only rebuilds when the graph data (`seq`) or theme changes. A same-data re-render (a selection click) just updates the selection + height and keeps the existing network and its viewport — so **zoom/pan are preserved**.
2. **Cache the layout for re-mounts.** When the component genuinely re-mounts (panel toggle changes the layout), node positions are restored from `localStorage` (keyed by `seq` + node set) with physics off — no re-stabilization, no flicker. Clicking **Render** bumps `seq` and recomputes a fresh layout.

Frontend only. JS syntax checked.

## Please test

- Zoom into the graph, then click a node - the zoom should stay put (no reset).
- Toggle the Details panel - the graph shouldn't flicker/re-layout.
- Click Render - it should recompute the layout.

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -6055,8 +6055,21 @@ def render_visualization():
             _component_path = _os.path.join(
                 _os.path.dirname(_os.path.abspath(__file__)), "lib", "graph_viewer"
             )
+            # Version the component name by a hash of its source, so any change to
+            # index.html serves under a fresh URL and browsers / the desktop
+            # webview can't run a stale cached copy (the webview's cache persists
+            # across launches via the storage_path added for #70).
+            import hashlib as _hashlib
+
+            try:
+                _gv_src = open(
+                    _os.path.join(_component_path, "index.html"), encoding="utf-8"
+                ).read()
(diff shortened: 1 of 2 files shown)


SuperCowProducts:
~~Everything looking good (did an octopus merge with the other PR)!~~

Just one minor glitch: after pressing `Save` on a node everything resets, which is understandable, as it causes a database update (there might be a way to mitigate this too, but it would require something like saving the zoom factor on the item last clicked on, so that auto zoom-in on it using the saved zoom factor can happen *after* a graph rebuild). Clicking on a node after that causes zoom out again though. Clicks subsequent to this have no negative impact until the next `Save`.

SuperCowProducts:
With latest commits it's saving everything after `Back to graph` 🎉

Issue #80, opened 2026-06-29
[idea] clicking on entity in Visualization allows editing it


SuperCowProducts:
Maybe it could bring up some kind of editing menu. In any case it would also conveniently double up as a way to copy entity data, which can be useful depending on what data is being stored (e.g. links / Markdown code blocks, which when rendered feature a handy button to copy contents). 

This is easier than exiting fullscreen mode, switching to `Classes`, selecting `Edit/Delete Class`, typing out the name of the class, making the edits, and going back to `Visualization` (along with similar workflow improvements for other types of entities which are already selectable like properties, individuals, relations, restrictions, annotations and concepts).

Now this may not end up being as fancy as [WebVOWL](https://github.com/VisualDataWeb/WebVOWL) (which includes an experimental Editing mode you can try out [here](https://service.tib.eu/webvowl/)), but I am sure it will be just as convenient (and will probably allow making more precise edits).

ralfbecher:
Shipped in v1.14.0. You can now interact with entities straight from the Visualization graph:

- A collapsible **Details side panel** (open it with the "‹" toggle on the right edge) shows the selected node's details and the IRI with a copy button, and lets you **edit a class inline** (name/label/comment/parent) without leaving the graph.
- Selecting a class also offers **"Open full editor"**, which jumps right into the editor with the class preselected, and a **"← Back to graph"** button to return.

Inline editing currently covers classes; the other entity types show details + open-in-editor and will get inline editing in a follow-up. Closing this as the core idea is in; thanks for the suggestion and all the testing.

PR #85, opened 2026-06-30
feat(viz): inline editing for properties and individuals in the panel

ralfbecher:
Follow-up to #80: extends the Details panel's inline editing from classes to the other entity types.

## What

When you select a node in the Visualization graph, the Details panel now shows a compact inline edit form for:

- **Object properties** — name, label, comment, domain, range (class).
- **Data properties** — name, label, comment, domain, range (datatype).
- **Individuals** — name, label, comment, add/remove class.

(Classes already had this.) Edges and unresolved nodes still show read-only details. The IRI (with copy button) now shows for all editable types.

## How

- New shared helpers `_apply_property_edit` and `_apply_individual_edit` (mirroring the existing edit forms; the class one is `_apply_class_edit`). They handle rename + field updates and return False on a rename clash.
- The panel body is extracted into `_render_panel_entity_editor`, which resolves the selected node to its entity (`_uid(uri)`) and renders the right form per type.

## Notes

- No version bump - batching for the next release.
- The existing list-tab edit forms still have their own inline logic; folding them onto these helpers is a reasonable future cleanup (left out here to keep the diff focused and low-risk).
- GUI behavior, so it needs a manual check.

## Please test

In the graph, select an object property / data property / individual, open the Details panel, edit a field, Save - the change should apply and the graph reflect it. The IRI copy button should appear for each.

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -1044,6 +1044,178 @@ def _apply_class_edit(ont, class_info, new_name, new_label, new_comment, new_par
     return True
 
 
+def _apply_property_edit(
+    ont, prop, new_name, new_label, new_comment, new_domain, new_range
+):
+    """Apply a property edit (rename + label/comment/domain/range).
+
+    Returns True on success, False (with an error message) on a rename clash.
+    Shared by the property edit forms and the Visualization side panel. The
+    caller handles the undo checkpoint and rerun. ``new_domain``/``new_range``
+    are passed straight to ``update_property`` (a class URI, a datatype name, or
+    "" to clear).
+    """
+    current_ref = prop["uri"]
+    if new_name and new_name != prop["name"]:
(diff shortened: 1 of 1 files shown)


SuperCowProducts:
Hi @ralfbecher, thank you for trying out the idea, it's already in pretty good shape. 

Only thing I am not getting is editable properties (was this PR also trying to do that?)

Also, out of curiosity, are there meant to be lines connecting annotations and triples to their respective entities? It appears there is intent for it to be so, but maybe it hasn't been implemented yet:
Screenshot_20260630_204111

SuperCowProducts:
So custom properties can definitely be edited with https://github.com/ralforion/orionbelt-ontology-builder/pull/85/changes/8ba6649294b900e364b671bb389c55c189113835 - hope I didn't miss that it was working before too: I was focusing on `subClassOf` and `disjointWith` which are unmodifiable anyway.

PR #92, opened 2026-07-11
fix(bulk): parse CSV without a header and support semicolon delimiter

### LINKED ISSUE #91: [bug] Classes > Bulk Operations: CSV not creating label and parent
If I create a class `Animal` and then  type

```
Dog, A Dog, Animal
```
into the `Class entries` box and preview with `Ctrl`+`Enter`, I get `Dog, A Dog, Animal` as one long `Name`. Is the line being parsed as CSV? If so it might not be using the standard delimiter `,`.

ralfbecher:
## Problem

Closes #91.

Bulk Add (Classes, Properties, Individuals) only treated input as CSV when the first line was a header row containing `name`. A user who typed a data line without the header, e.g.

```
Dog, A Dog, Animal
```

got the whole string parsed as one long class `Name` instead of `name` / `label` / `parent`. The comma was the correct delimiter; the parser simply required an explicit `Name, Label, Parent` header before it would split rows.

## Changes

`parse_bulk_text` now:

- **Header-optional:** accepts `default_columns` and maps a delimited data line positionally onto them when no header row is present. Fixes the reported case.
- **Semicolon delimiter:** if any line contains a `;`, `;` is used as the delimiter (otherwise `,`), so a label can legitimately contain a comma, e.g. `Dog; A Dog, the domestic one; Animal`.

The three bulk-add pages pass their column sets (`name,label,parent` / `name,domain,range,label` / `name,class,label`), and the captions note that `;` can be used when a label contains commas.

Backward compatible: without `default_columns` and no header, a comma line still stays a single name (documented simple-mode behavior); an explicit header still takes precedence over defaults.

## Tests

Added 6 cases to `tests/test_bulk.py`: header-less comma (the exact issue), header-less semicolon, semicolon preserving a comma in a label, semicolon header row, simple-mode fallback without defaults, and header-precedence over defaults.

Full suite: 336 passed. Ruff check and format clean.

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -1780,16 +1780,19 @@ def render_classes():
         if bulk_op == "Add":
             st.subheader("Bulk Add Classes")
             st.caption(
-                "Enter one class name per line, or use CSV format: Name, Label, Parent"
+                "Enter one class name per line, or use CSV format: Name, Label, "
+                "Parent. Use ';' as the separator when a label contains commas."
             )
             bulk_text = st.text_area(
                 "Class entries",
                 height=200,
                 key="bulk_classes_text",
-                placeholder="Dog\nCat\nBird\n\nor CSV:\nName, Label, Parent\nDog, A Dog, Animal\nCat, A Cat, Animal",
+                placeholder="Dog\nCat\nBird\n\nor CSV:\nDog, A Dog, Animal\nCat, A Cat, Animal",
             )
             if bulk_text:
-                entries = ont.parse_bulk_text(bulk_text)
+                entries = ont.parse_bulk_text(
+                    bulk_text, default_columns=["name", "label", "parent"]
+                )
                 if entries:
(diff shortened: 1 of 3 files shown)


ralfbecher:
Addressed the semicolon auto-detection regression (P2) in 0730abe.

The delimiter is no longer chosen by "semicolon anywhere". It is now taken from the first line that contains a delimiter, picking whichever of `;` / `,` is more frequent on that single line. So a comma CSV whose header uses commas stays comma-delimited even when a later row's label contains a semicolon, and the reverse holds for semicolon CSVs.

Verified against the reported case:

```
parse_bulk_text("Name, Label, Parent\nDog, A dog; domestic, Animal",
                default_columns=["name", "label", "parent"])
# -> [{"name": "Dog", "label": "A dog; domestic", "parent": "Animal"}]
```

Added two regression tests (`test_semicolon_in_label_does_not_flip_comma_csv`, `test_semicolon_in_label_header_less_comma_csv`). Full suite: 338 passed, ruff clean.

Issue #90, opened 2026-07-07
[idea] Desktop App Title matches ontology name


SuperCowProducts:
I recently suggested https://github.com/ralforion/orionbelt-ontology-builder/issues/87, but realised it would be a huge effort to implement when it is already possible to open multiple instance of OrionBelt anyway. This simpler idea came to my mind when dealing with these multiple instances, as when switching around tabs all you can see is the filename and you can easily create / delete entities from the wrong ontology which can be annoying.

I wonder if Streamlit is on our side in this case 🤔

SuperCowProducts:
Hi @ralfbecher, thanks for providing it on the browser! Can the same feature work for the desktop app too?

PR #102, opened 2026-07-12
feat(desktop): native window title follows the ontology name (#90)

ralfbecher:
## What

The browser tab already shows the ontology name (#90/#101), but the **native desktop window** kept its static launch title (`OrionBelt Ontology Builder`), because pywebview does not propagate the page's `document.title` to the OS window.

This makes the native window title track the ontology name too, so multiple desktop instances are distinguishable.

## How

Hook `webview.create_window` (same monkeypatch pattern the launcher already uses for `webview.start`) so each window:
1. **Exposes a title setter** to JS (`orionbelt_set_window_title`) that calls `window.set_title(...)`.
2. **On load, injects a MutationObserver** on the page `<title>` that pushes `document.title` (which `st.set_page_config` keeps set to the ontology name) back through that setter.

The setter ignores the Streamlit default / empty title so the window keeps a meaningful name during the brief pre-connect window. Everything is wrapped defensively: any failure leaves the static title in place rather than breaking the desktop launch, and it no-ops entirely when `create_window` is absent (e.g. a stubbed webview in tests).

## Verification

- Unit tests: the hook exposes the setter and injects the observer; the setter mirrors a real title and falls back to the app name for the Streamlit default/empty; and it no-ops without `create_window`. (14 desktop tests, 419 total, ruff clean.)
- Smoke-tested against **real** pywebview: confirmed `Window.events.loaded += handler`, `Window.expose(func)`, `set_title`, and `evaluate_js` all exist and are used correctly.

**What I could not verify:** the actual native window title updating requires a real display + GUI event loop, which I don't have in this environment. Please run `orionbelt-ontology-builder-desktop` and confirm the window title shows the ontology name (and updates when you set the ontology's Label on the Dashboard). The change is defensive, so worst case the title stays static — it can't break the launch.

### DIFF
--- a/orionbelt_ontology_builder/desktop.py
+++ b/orionbelt_ontology_builder/desktop.py
@@ -52,6 +52,81 @@ def _preferred_gui() -> str | None:
     return None
 
 
+# Injected into the desktop webview: mirror the page's document.title (set by
+# st.set_page_config, so it carries the current ontology name) onto the native
+# window through the exposed bridge below. pywebview does not propagate
+# document.title to the OS window on its own, so the window would otherwise keep
+# its static launch title (issue #90).
+_TITLE_SYNC_JS = """
+(function () {
+  if (window.__orionbeltTitleSync) return;
+  window.__orionbeltTitleSync = true;
+  function push() {
+    try {
+      var api = window.pywebview && window.pywebview.api;
+      if (api && api.orionbelt_set_window_title) {
+        api.orionbelt_set_window_title(document.title);
+      }
+    } catch (e) {}
+  }
+  var head = document.head || document.documentElement;
+  new MutationObserver(push).observe(head, {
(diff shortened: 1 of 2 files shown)


ralfbecher:
Verified live against real pywebview (macOS/Cocoa), driving `document.title` changes through the actual shipped hook (`_install_window_title_sync`, real exposed setter, real MutationObserver JS):

- Initial load pushed `Initial · OrionBelt`.
- Changing `document.title` to `Pizza Ontology · OrionBelt` fired the observer and updated the native title.
- A second change to `Second Change · OrionBelt` did the same.

Result recorded at the native `set_title` boundary: `["Initial · OrionBelt", "Pizza Ontology · OrionBelt", "Second Change · OrionBelt"]`, with no `set_title` errors. So the title updates on ongoing changes (e.g. renaming the ontology Label), not just at load.

PR #199, opened 2026-07-27
feat: paste a class list into the Visualization filter

### LINKED ISSUE #179: [idea] allow pasting names of multiple classes in `Filter Classes`
Cool for when you want to restore a specific visualization of the graph with only the key classes included, or pretty much anything. Space between classes is sufficient right? We would need a way of specifying the namespace of classes with the same name (e.g. `fn:zero`), currently both classes with same name are displated.

ralfbecher:
Closes #179

## What

The Visualization "Filter Classes" control could only be driven one click at a time, and it listed raw local names, so two classes sharing a name (`fn:zero` and `math:zero`) showed up as two identical entries that toggled together.

**Namespace disambiguation.** The filter options are now the namespace-tagged display names the rest of the app already uses (`zero (fn)` / `zero (math)`), and the graph builder resolves the selection to URIs, so hiding one of two same-named classes hides only that one. The Find-entity picker beside it keyed on the bare name too and had the same collapse, so it moved to the same display names.

**Pasting.** A text area under the multiselect takes a pasted class list and applies it in one click. Tokens separate on whitespace, commas, semicolons or line breaks, and each may be:

- a plain local name: `Person` (an ambiguous name selects every namespace carrying it)
- a prefixed name: `fn:zero` (picks exactly one)
- the displayed form: `zero (fn)`
- a full URI, optionally in `<>` or quotes

Matching is exact first, then case-insensitive. Tokens that match nothing are reported in a warning instead of being silently dropped.

The panel also prints the current selection back in that same syntax, so a narrowed view can be copied and pasted back later. That was the use case in the issue: restoring a specific visualization with only the key classes included.

## Testing

- `tests/test_viz_class_paste.py`: 18 tests covering entry building, token round-trip, every separator and token form, ordering, dedup, case handling, unmatched reporting, and that a pasted selection survives the next render's reconciliation.
- Full suite (670), ruff format/check and mypy all pass.
- Verified live against an ontology with `fn:zero` + `math:zero`: pasting `Person, fn:zero\nNonsense` narrows the graph to exactly `Person` and `zero (fn)`, warns about `Nonsense`, and the printed tokens read `Person fn:zero`. Find-and-centre on `zero (math)` reveals only that class, and focus mode still seeds from the selection.

## Note

While rewiring the class filter I corrected one adjacent line: in the "Triples" view, class subjects were mapped to `cls["name"]` as their node id, but class nodes are keyed by a URI hash, so those triple edges pointed at nonexistent nodes. The same defect exists for individuals (`ind_{name}` vs `ind_{uid}`) at two other spots; I left those alone as unrelated scope and will file them separately.

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -6683,10 +6683,10 @@ def build_class_hierarchy_text(classes):
     return "\n".join(lines)
 
 
-def reconcile_class_filter(all_class_names, selected, known, replaced=False):
+def reconcile_class_filter(all_class_uris, selected, known, replaced=False):
     """Reconcile the Visualization "Filter Classes" selection with the ontology.
 
-    Diffs the current class names against those seen on the previous render
+    Diffs the current class URIs against those seen on the previous render
     (``known``) rather than resetting on every edit, so adding a class or a
     restriction no longer wipes a narrowed filter (issue #180):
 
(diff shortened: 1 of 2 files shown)


SuperCowProducts:
Awesome! It only failed with an awkwardly-named class I had (`'rawdict'`) - not sure if `'` would actually be allowed in a class name honestly, but if they are just something to keep in mind.

ralfbecher:
@SuperCowProducts yes, not allowed

SuperCowProducts:
@ralfbecher now I realise what happened: I created that class in a previous version which didn't check for the illegal character `'`, although current version definitely stops this, so all good :+1:

PR #202, opened 2026-07-27
feat: filter individuals, in a filter panel that stops growing

### LINKED ISSUE #196: [idea] allow individuals in `Filter Classes` / `Find and centre ...`

ralfbecher:
Closes #196

## What

The Visualization filter only covered classes. Individuals were all-or-nothing via the entity-type checkbox, and while Find-and-centre already listed them, it could never reveal a hidden one because nothing could hide one.

## The panel shape

Rather than stack a second block of controls under the first, the panel edits one kind at a time, picked by a segmented control. The height is constant however many kinds exist, so a future Data Properties or SKOS filter costs one segment rather than five more rows.

Before, nine rows:

```
▼ Filter Classes
  ☐ Focus on one node
  Select classes to display
  [Person ×] [Organization ×]          ▾
  ( Select all classes )
  Or paste a class list
  [                                    ]
  [                                    ]
  ( Apply pasted list )
  Current selection — copy to restore it later:
  ┌────────────────────────────────────┐
  │ Person Organization                │
  └────────────────────────────────────┘
```

After, five, with the new filter in it:

```
▼ Filter Nodes
  ☐ Focus on one node
  [ Classes │ Individuals 2/5 ]
  [alice ×] [acme ×]                   ▾
  ( Select all )   ( Paste / copy )
```

Measured in the browser: 265px, versus 303px for the class-only panel it replaces.

What buys the room is moving the paste/copy affordance added in #179 into a popover behind one button. The trade-off is that the copyable token string is now a click away rather than on screen; it is a save/restore affordance rather than something you read at a glance, so that seems the right place for it.

Each segment shows a "shown/total" count when that kind is narrowed, so a filter applied to a kind you are not currently looking at stays visible. Only kinds whose entity-type checkbox is on are offered, since filtering something that is not drawn is meaningless.

## Implementation

The filter helpers were already generic over `{name, uri}` entries and never looked at anything class-specific, so they only needed renaming: `build_class_filter_entries` -> `build_filter_entries`, `parse_class_filter_text` -> `parse_filter_text`, `class_filter_token` -> `filter_entry_token`, `reconcile_class_filter` -> `reconcile_filter_selection`.

Per-kind state lives in one dict keyed by kind rather than parallel variables per kind, driven by a `_FILTER_KINDS` descriptor list. The Find-reveal branch walks those kinds instead of special-casing `Class`. Individual labels are namespace-disambiguated the same way class labels are, so two individuals sharing a local name are separately selectable.

The segmented control uses stable kind keys as its options with the count rendered via `format_func`. That matters: with the count baked into the option strings, a narrowing rewrites the stored widget value out of existence and the control breaks. I confirmed that failure mode before settling on `format_func`.

Because #201 made every edge gate on the nodes actually emitted, hiding an individual takes its type links, object-property assertions and annotations with it for free.

## Testing

- Four new cases in `tests/test_viz_graph_edges.py`: individuals shown by default, hiding one removes only that node and leaves no dangling edges, hiding all of them leaves the classes intact, and the two filters are independent.
- 683 tests pass, ruff format/check and mypy clean.
- Verified live end to end: switching segments, narrowing by multiselect, pasting `alice, acme\nnobody` (applies both, reports `Ignored, no such individual: nobody`), Select all, and Find-and-centre on the hidden `carol`, which reveals her, adds her to the graph and centres the camera on her node. Focus mode still builds the full graph and prunes, ignoring both filters as before. Zero dangling edges throughout.

Note: `st.segmented_control` cannot be driven through `AppTest` (its test harness iterates the value as a list and fails), so the segment switching itself is covered by the live check rather than a unit test. The filter state it selects between is unit-tested.

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -6683,44 +6683,70 @@ def build_class_hierarchy_text(classes):
     return "\n".join(lines)
 
 
-def reconcile_class_filter(all_class_uris, selected, known, replaced=False):
-    """Reconcile the Visualization "Filter Classes" selection with the ontology.
+# The node kinds the Visualization filter can narrow. One entry per kind is all
+# a new filter needs: the controls are rendered from this list, so adding a kind
+# costs a segment in the selector rather than another block of UI (issue #196).
+# ``toggle`` is the entity-type checkbox that has to be on for the kind to be in
+# the graph at all; a kind whose toggle is off is not offered. ``singular`` is
+# the prefix the Find-and-centre picker labels this kind with ("Class: Person").
+_FILTER_KINDS = (
+    {
+        "key": "class",
+        "toggle": "show_classes",
+        "label": "Classes",
+        "singular": "Class",
+        "noun": "class",
+        "plural": "classes",
+    },
+    {
+        "key": "ind",
+        "toggle": "show_individuals",
+        "label": "Individuals",
+        "singular": "Individual",
+        "noun": "individual",
+        "plural": "individuals",
+    },
+)
 
-    Diffs the current class URIs against those seen on the previous render
-    (``known``) rather than resetting on every edit, so adding a class or a
-    restriction no longer wipes a narrowed filter (issue #180):
 
-    - classes deleted since last render drop out of the selection;
-    - classes created since last render are added (new content is shown by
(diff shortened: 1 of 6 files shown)


SuperCowProducts:
Very neat design

Issue #180, opened 2026-07-24
[bug] `Filter Classes` deleted when class is created


SuperCowProducts:
I have a feeling this may be defensive programming preventing visualization of classes that don't exist anymore or have been renamed, although it would be nice if only the ones that were deleted were removed (and have the ones renamed from `Classes` renamed in `Filter Classes` too.

[UPDATE] Same thing happens after adding a restriction - note these tests were performed with `Focus on one node` disabled

ralfbecher:
Fixed in #183, released in [v1.17.3](https://github.com/ralforion/orionbelt-ontology-builder/releases/tag/v1.17.3) and live on the [hosted demo](https://orionbelt.streamlit.app).

The `Filter Classes` selection is now kept when a class or restriction is added, instead of being cleared.

PR #212, opened 2026-07-30
feat: persist node filters and focus seeds per linked file (#164)

### LINKED ISSUE #164: [desktop] [idea] save 'focus seeds'
https://github.com/ralforion/orionbelt-ontology-builder/pull/163 mentions focus seeds are not persisted due to possibility of opening a different ontology, but with the desktop app we have the advantage of being able to track a specific ontology (file path is already saved in `config.json`): therefore, so long as `linked_file` exists, I think we would also be able to load saved focus seeds for it without any issues!

ralfbecher:
Closes #164.

## Why

PR #163 persisted the Visualization display settings across sessions, but deliberately left out two: the node filters and the focus seeds. Both name specific entities, so restoring `Class: Person` into a different ontology would be meaningless.

As @SuperCowProducts pointed out on the issue, that objection does not hold on desktop: the linked working file identifies the ontology, and its path already lives in `config.json`. So this state can be saved against that path and restored while the same file is still linked.

Scope confirmed with the reporter: both the node filters and the focus seeds, not just the seeds. They are coupled anyway, since the focus seeds default to whatever is selected in the class filter, so restoring one without the other would leave the graph half restored.

## What is stored

A new `viz_file_state` entry in `config.json`, keyed by linked path:

```json
"viz_file_state": {
  "/home/me/pizza.owl": {
    "hidden_class_uris": ["http://...#Veneziana"],
    "focus_seeds": ["Class: American"]
  }
}
```

It records what the user **hid**, not what is selected. Everything is shown by default, so the hidden set is normally empty or tiny, which matters for an ontology with thousands of entities.

Storing the hidden set also restores identically to storing the full selection. The saved state is fed *through* `reconcile_filter_selection` rather than around it, so the existing rules still apply:

- an entity added to the file since the last session is absent from the hidden set, so it shows up, the way new content always does;
- an entity deleted since drops out;
- everything else keeps the state the user left it in.

Focus seeds are stored as their display labels, which is what session state already holds. The focus-mode block already prunes seeds the ontology no longer has and falls back to the first available node, so a file edited outside the app degrades rather than breaking.

## Scope and limits

- Desktop and local mode only, gated on `local_persist_enabled()` plus a linked path. A cloud session has neither and keeps today's per-session reset. That is a deliberate behavior difference between the two.
- "Same path" is a proxy for "same ontology", not a guarantee. Someone can edit the linked file outside the app; the prune above is what handles that.
- `config.json` growth is bounded to the 20 most recently changed files, oldest evicted first.
- Saving no-ops when nothing changed since the last render and when the state on disk already matches, so ordinary reruns do not rewrite the file.

## Testing

- New `tests/test_viz_file_state.py`, 19 tests: what the payload stores, the disk round trip, entities added or deleted from the file since, scoping to the linked path, relinking mid-session, restore running once per file, the cloud and no-linked-file cases, the eviction cap, the no-rewrite guard, and junk on disk.
- Full suite: 712 passed. ruff 0.16.0 format and check clean, mypy clean.
- Verified live in local-persist mode against an isolated `HOME`, so no real user config was touched. Linked `pizza.owl`, hid a class, turned focus mode on, confirmed `config.json`, then restarted the server: the hidden class was still hidden (98 of 99 chips, `Veneziana` absent) and all 98 focus seeds came back.

Note for reviewers: the class filter multiselect is not rendered while focus mode is on, so the filter restore has to be checked with focus mode off.

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -865,6 +865,15 @@ def _clear_viz_file_session_state() -> None:
         st.session_state.pop(f"_viz_cfg_known_{kind['key']}_uris", None)
     st.session_state.pop("_viz_cfg_focus_seeds", None)
     st.session_state.pop("_viz_pending_focus_seed_ids", None)
+    # The mutation counters seen on the last render belong to the file we just
+    # left, and loading the new one bumps the ontology's counter without a
+    # matching edit. Left in place they would read as "the ontology was
+    # replaced", which resets the filter to everything shown and so discards the
+    # state we are about to restore. Nothing can leak across the switch anyway:
+    # the selection itself was just cleared, so the only thing the reconcile has
+    # to diff against is the new file's own saved state.
+    st.session_state.pop("_viz_cfg_seen_mutation", None)
+    st.session_state.pop("_viz_cfg_seen_edits", None)
 
 
 def _restore_viz_file_state() -> dict:
@@ -889,6 +898,40 @@ def _restore_viz_file_state() -> dict:
     store = local_store.load_config().get(VIZ_FILE_STATE_KEY)
     entry = store.get(file_id) if isinstance(store, dict) else None
     return entry if isinstance(entry, dict) else {}
+
+
+def viz_ontology_was_replaced() -> bool:
+    """Whether the whole ontology was swapped since the last graph render.
+
+    A mutation-counter jump not matched by the edit counter means a
+    load/import/new/undo rather than an incremental edit, so the node filters
+    reset to "everything shown" instead of diffing against an ontology that may
+    reuse URIs for unrelated entities.
+
+    Must be evaluated *after* :func:`_restore_viz_file_state`, which clears the
+    counters when the linked file changed: loading the new file bumps the
+    ontology counter, and treating that as a replacement would throw away the
+    state being restored for it (issue #164).
+    """
+    mutation = st.session_state.get("_ont_mutation_count", 0)
+    edits = st.session_state.get("_ont_edit_count", 0)
+    prev_mutation = st.session_state.get("_viz_cfg_seen_mutation")
+    prev_edits = st.session_state.get("_viz_cfg_seen_edits", 0)
+    return prev_mutation is not None and (mutation - prev_mutation) != (
+        edits - prev_edits
+    )
+
+
+def viz_mark_ontology_seen() -> None:
+    """Record the counters this render reconciled against.
+
+    The counterpart to :func:`viz_ontology_was_replaced`, which compares the
+    next render's counters with these.
+    """
+    st.session_state["_viz_cfg_seen_mutation"] = st.session_state.get(
+        "_ont_mutation_count", 0
+    )
+    st.session_state["_viz_cfg_seen_edits"] = st.session_state.get("_ont_edit_count", 0)
 
 
 def seed_filter_from_saved(all_uris, hidden, selected, known):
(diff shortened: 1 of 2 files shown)


ralfbecher:
Thanks, all three confirmed and fixed in 8dfc0ba.

**P1** was slightly worse than described. The stale `_viz_cfg_focus_seeds` guard means it is not only that the new file's state is ignored and erased: the *previous* file's seeds are still in session, so they get written back out under the *new* file's key. Cross-contamination, not just erasure.

Fixed by clearing the per-ontology session keys (`_viz_cfg_selected_*_uris`, `_viz_cfg_known_*_uris`, `_viz_cfg_focus_seeds`) when the linked file id changes, in a new `_clear_viz_file_session_state()`. Two tests added: one asserting the stale keys are gone and the new file's hidden set then applies, one asserting the old file's seeds are not written under the new file's key.

Note on the test you suggested: my first version of it persisted before ever restoring, and failed. That sequence cannot happen in the app, since `render_visualization` always restores before it persists, so I made the test mirror that order rather than loosen the code. The clearing is deliberately scoped to a genuine switch, not the first render.

**P2** fixed: the fingerprint is now recorded only after `save_config()` returns, and the `except OSError` returns early so a later rerun retries. Test drives a failing `save_config`, asserts no fingerprint is recorded, then asserts the same unchanged state is written once the disk recovers.

**P3** fixed: seeds are stored as the graph's node ids rather than display labels. For classes, individuals and data properties those derive from the URI, so a label that gains a namespace tag still resolves. The ids are mapped back to labels on render, after `focus_targets` is built. Tests cover the id round trip and the disambiguation case.

One residual limitation worth recording: SKOS concept node ids are `skos_{name}`, so a *renamed* concept still loses its seed. That is the graph builder's own id scheme rather than something this PR introduces, and the seed simply drops out as before.

Verified live again in local-persist mode against an isolated `HOME`:
- Seeds round-trip as `_uid()` ids across a server restart (`Caprina` and `Margherita` came back, class filter still 2 of 99).
- Relinking to another ontology mid-session now shows all 15 of its classes, leaks no seeds, and writes no entry for it, with the first file's entry left intact.

718 tests pass (25 in `test_viz_file_state.py`), ruff and mypy clean.

ralfbecher:
Confirmed and fixed in 5501c15. Good catch, and my previous live check could not have caught it: I relinked to a file with *nothing* saved, so "everything shown" was indistinguishable from a correct restore.

The counters do survive the switch, `_load_linked_file` bumps the ontology counter with no matching edit, so `replaced` was still true and the reconcile threw away the hidden set that had just been restored. Fixed by clearing `_viz_cfg_seen_mutation` / `_viz_cfg_seen_edits` alongside the selection. That is safe rather than papering over the guard: the selection was cleared in the same step, so there is nothing from the previous ontology left for the reconcile to inherit, and the only thing it diffs against is the new file's own saved state.

One thing your report surfaced indirectly: `replaced` was computed *before* `_restore_viz_file_state()` ran, so the ordering was load-bearing but invisible. It is now `viz_ontology_was_replaced()`, called after the restore, with the dependency stated in its docstring, plus `viz_mark_ontology_seen()` as the counterpart that records the counters. Extracting it is also what makes the regression testable.

Three tests added, and I verified the first one fails without the fix rather than passing vacuously:
- relinking mid-session no longer reads as a replacement, and the new file's hidden set survives the reconcile;
- a genuine swap over the *same* linked file (import / new / undo) still resets the filter, so the #180 behavior is intact;
- an ordinary edit is still not a replacement.

Verified live in local-persist mode, this time relinking to a file that *does* have saved state: pizza came back with its 97 hidden classes (2 of 99 shown) instead of all 99, and its `config.json` entry was left intact rather than overwritten.

721 tests pass, ruff and mypy clean.

ralfbecher:
Real leak, confirmed and fixed in 7baf287. The proposed fix location does not close it, though, and I only found that out by trying it.

`Apply Import` goes through `save_checkpoint()` (app.py:6101), which bumps the edit counter alongside the mutation counter. So an import is deliberately *not* a replacement by `viz_ontology_was_replaced()` — that is the #180 design, where the gap between the two counters is what separates an edit from a load/new/undo. I implemented the suggested `if replaced: drop seeds`, ran the exact scenario live, and the leak was still there: with onto1 (Person, Dog) linked and both seeded, importing onto2 (Person, Cat) over it still persisted onto2's `Person` as the seed.

The real problem is the one your finding points at underneath: seeds are pruned by label, and a label does not identify an entity. So the fix compares each seed label against the node id it resolved to on the previous render (`prune_reused_focus_seeds`), which catches the swap whatever caused it — import, relink, undo — without depending on the counters at all. A label with no recorded id is one the user just picked, so it survives.

Live result on that same scenario:

| | seeds persisted after importing onto2 |
| --- | --- |
| before | `['onto2 Person']` — stale label re-pointed at a different entity |
| after | `['onto2 Cat']` — both stale seeds dropped, falls back to the first node |

The fallback to the first node is the existing behavior whenever no seed survives; I did not change it.

Five tests added: a reused label naming a different entity is dropped, a label still naming the same entity is kept (including when a namespace tag appears next to it), a freshly picked seed with no recorded id survives, pruning is a no-op before anything is recorded, and a dropped seed is not persisted.

727 tests pass, ruff and mypy clean.

PR #218, opened 2026-08-02
fix: let focus mode reach nodes past the graph's render cap (#216)

### LINKED ISSUE #216: [bug] `Focus on one node` does not work for some ontology items
In a large ontology, I have found certain nodes (`trig`, `trig-fn`) don't work with the `Focus on one node` toggle. I have tried clearing `.orionbelt-ontology-builder` but to no avail, so maybe there is something wrong with [my file](https://paste.sr.ht/~handymanny/70214e32103692b13dbb5b75f214f503a1f3e11a)?
<details>

<summary>
original
</summary>

What I did to fix is cleared the `focus_seed_ids` array. Only happened once after a lot of day-to-day usage, so may have just been a temporary glitch.

</details>

[UPDATE] A cool little workaround in this case is to tick the `Triples` checkbox, although I do wonder how I could debug this, it does seem rather mysterious for now 🤔 

Also curious to see deleting and recreating `trig` doesn't seem to fix this; clearing the ontology and creating `trig` and `trig-fn` items solves the problem, so I may be able to pinpoint a hypothetical corrupt triple which is stopping perhaps multiple classes from displaying their any of their connections.

ralfbecher:
Closes #216

## What was wrong

The graph draws at most 500 nodes, which is what keeps vis-network (and the desktop webview) responsive. That cap was applied while **assembling** the graph — but focus mode assembles everything and prunes to the seeds' neighbourhood afterwards, so a class past the cap never got a node at all. The seed was then not found, and focus deliberately drew nothing.

Classes are built in alphabetical order, so which ones this hit came down to their name. In @SuperCowProducts' ontology (539 classes, thanks for the file, it made this quick to pin down) `trig` is the 506th, so it and everything after it were unreachable, while earlier names worked. That also explains the rest of the report: deleting and recreating `trig` puts it back in the same alphabetical spot, and a fresh ontology with only `trig` and `trig-fn` is nowhere near the cap.

Reproduced with the attached ontology: focusing on `trig` built 0 nodes, on `trig-fn` 0 nodes, while `A` and `opt` built theirs fine.

## What changed

- **The assembly allowance is raised for focus mode**, where the browser only ever sees the pruned result. Assembly is plain dicts and stays well under a second at 5000 nodes.
- **The prune holds the render cap instead**, expanding ring by ring from the seeds and stopping when the budget is spent. The seeds alone can overflow it, since they default to every selected class.
- **The allowance is tied to the prune actually running**, not to the mode being on. With focus on and every seed cleared nothing prunes, and the browser would otherwise be handed the whole ontology.
- **The graph now says what it left out.** None of this was visible before: nodes dropped by the cap went missing silently, taking every edge that needed them, which reads as unrelated classes losing their connections. An empty focus now explains itself rather than looking broken.

The Visualization tab picker also seeds its selection in `session_state` rather than passing `default=`, matching the Classes, Relations and Restrictions pages.

## Testing

- `tests/test_viz_node_cap.py`: the regression itself (a class past the cap focuses and keeps its neighbours), the cap still holding without focus, a neighbourhood and a seed set larger than the cap being cut with no dangling edges, and a seed past the assembly cap explaining the empty graph.
- Full suite, ruff and mypy pass.
- Verified in the running app with the reported ontology: focusing on `trig` returns its neighbourhood, a plain render still stops at 500 nodes and now says "39 of 539 classes are not drawn", and a focus over every class is bounded to exactly 500 with a notice. Driving it live is what caught two holes in the first version of this fix.

## Not changed

Focus seeds default to *every* selected class, so ticking "Focus on one node" starts by focusing on all of them. That is worth revisiting, but it affects every node equally and is not what made specific ones fail.

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -2378,6 +2378,27 @@ def render_dashboard():
 # every per-item list view shares this cap (issue #146).
 LIST_PAGE_SIZE = 50
 
+# How many nodes the Visualization graph may draw. vis-network (and the desktop
+# webview especially) bogs down well before a large ontology's worth of them.
+GRAPH_MAX_NODES = 500
+
+# How many it may *assemble* in focus mode, where the drawn graph is the pruned
+# neighbourhood of the seeds rather than everything built. Assembly is cheap —
+# plain dicts — so this can be far larger than what is drawn, which is what lets
+# a class deep into a big ontology be focused on at all (issue #216).
+FOCUS_BUILD_MAX_NODES = 5000
+
+
+def graph_node_cap(focus_pruning: bool) -> int:
+    """How many nodes a graph build may assemble (issue #216).
+
+    Going past what can be drawn is only safe because focus mode prunes back to
+    the seeds' neighbourhood afterwards, so the allowance is tied to that prune
+    actually running — not to focus mode being on. With the mode on but every
+    seed cleared nothing prunes, and the browser would be handed the lot.
+    """
+    return FOCUS_BUILD_MAX_NODES if focus_pruning else GRAPH_MAX_NODES
+
 
 def _page_bounds(total: int, page: int, page_size: int) -> tuple[int, int, int, int]:
     """Resolve a 1-based ``page`` over ``total`` items into slice bounds.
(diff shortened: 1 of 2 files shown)


SuperCowProducts:
Thanks very much - it was a particularly important change for working with large ontologies!

Issue #221, opened 2026-08-03
[idea] add entities from visualisation


SuperCowProducts:
OrionBelt has come a long way in terms of visualisation-backed workflow, and it is now possible to edit all sorts of entities directly from visualization. This is why I thought it would be neat to be able to add entities (e.g. classes, individuals, relations, restrictions) directly from visualization. This opens up opportunities compared to text-based method of adding classes. 

For example, selecting a pre-existing class before creating a new class could automatically make the new class a subclass of the selected class. Or maybe, if the entity being created is a `subClassOf` relation it would have the opposite effect (the selected class would become a subclass of another class which can be selected on the graph).

ralfbecher:
Thanks, this is a natural next step after #152 / #217.

I had a look at what it would take. The details panel already resolves any selected node or edge to its editor, so hosting an add form there is mostly plumbing. The interesting part is your second point: how the graph supplies the *other* endpoint when the thing being created is a relation or a restriction.

Two ways to do it:

**1. Explicit pick step.** Select a class, open "Add relation", press a "Pick on graph" button, then click the second node. Unambiguous, but it costs an extra click and needs a change to the graph component, since Ctrl/Cmd-click is already taken by focus.

**2. Remember the previous selection.** No mode, no button: you click A, then click B, and the add-relation form opens prefilled with B as subject and A as object. Both endpoints stay editable dropdowns, so the pair is a suggestion you can override before anything is written. Zero extra clicks, and no component change at all.

My instinct is 2, with the prefill labelled clearly so it never quietly asserts a triple you did not mean. The risk is the mirror image of the benefit: clicking around the graph to read it leaves a stale "previous" node sitting in the form.

The single-selection cases (new class as a subclass of the selected one, new individual of it, new property with it as domain, restriction on it) need one click either way, so this only really decides the two-endpoint cases.

Does 2 match how you would actually use it, or would you rather have the explicit pick step? And which additions would you reach for most? That would tell me what to build first.

SuperCowProducts:
[...]

PR #237, opened 2026-08-07
feat: add a restriction by clicking the two classes it relates (#221)

ralfbecher:
Third step of #221, after #230 (classes) and #231 (relations). Completes the set the reporter named as daily use.

Same arm-then-pick as the relation form, reusing `resolve_picked_object`: press **Add restriction** with a class selected and it becomes the class the restriction applies to; the next class you click is the one it points at.

### Which types are offered

Only the ones that actually have a second class:

| type | what the picked class fills |
| --- | --- |
| `someValuesFrom`, `allValuesFrom` | the value |
| `minQualifiedCardinality`, `maxQualifiedCardinality`, `qualifiedCardinality` | the `owl:onClass`, plus a count |

`hasValue` and the plain cardinalities describe one class on its own, so the graph has no second end to offer and they stay on the Restrictions page. The field label changes to say which slot the picked class is filling, and the count appears only for the qualified ones.

### A pre-existing bug, fixed by sharing the classification

The page decided whether a type was qualified with:

```python
if "Qualified" in restriction_type:
```

That matches `minQualifiedCardinality` and `maxQualifiedCardinality`, but **not** `qualifiedCardinality`, whose q is lower case. So that one type never got an onClass selector, and the page wrote:

```turtle
:Person rdfs:subClassOf [ a owl:Restriction ;
        owl:onProperty :hasRole ;
        owl:qualifiedCardinality "2"^^xsd:nonNegativeInteger ] .
```

A qualified cardinality with no `owl:onClass`, which is not valid OWL 2 and which no reasoner will accept. I needed this classification for the panel anyway, so it is now one case-insensitive helper used by both callers, which fixes the page as a consequence. A test asserts the classifiers cover every type the engine knows, so a new type cannot be silently skipped.

### Why the type selector is outside the form

A Streamlit form batches until submit and does not rerun when a widget inside it changes. With the type inside, switching to a qualified type left the label reading "Value (Class)" and never rendered the count field, so pressing Add submitted the new type against the old type's shape. Verified that failing, then moved the selector out.

Worth noting the **Restrictions page has the same shape** for its own value fields, so the value widget there can lag the chosen type by one submit. That is pre-existing and untouched here; happy to fix separately if you want it.

### Verified live

| step | result |
| --- | --- |
| select `Person`, press Add restriction | holds at "Restriction on Person / Now click the class it restricts to" |
| click `Role` | form opens, Apply to = Person, Value (Class) = Role, no count field |
| switch type to `qualifiedCardinality` | label becomes "Qualified on Class", a Cardinality field appears |
| submit | writes `owl:onClass :Role` with `owl:qualifiedCardinality "1"`, confirmed in Source |
| Restrictions page, add a `someValuesFrom` | still works, listed alongside |

821 tests pass, ruff and mypy clean.

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -2484,7 +2484,7 @@ def _render_panel_add_restriction_button(classes, ntype, ename):
         st.rerun()
 
 
-def _render_panel_add_restriction_form(ont, classes, properties, ntype, ename):
+def _render_panel_add_restriction_form(ont, classes, object_props, ntype, ename):
     """Add a restriction by clicking the two classes it relates (issue #221).
 
     Same arm-then-pick as the relation form: the selected class is the one the
@@ -2493,15 +2493,26 @@ def _render_panel_add_restriction_form(ont, classes, properties, ntype, ename):
     value restrictions and the three qualified cardinalities. hasValue and the
     plain cardinalities say something about one class on its own, so the graph
     has no second end to offer and they stay on the Restrictions page.
+
+    ``object_props`` only, deliberately. Everything this flow can produce fills a
+    class slot: the value of someValuesFrom/allValuesFrom, or the owl:onClass of
+    a qualified cardinality. On a data property those slots take a data range
+    instead, so offering one here would write ``owl:someValuesFrom :SomeClass``
+    against an ``owl:DatatypeProperty`` — an axiom no reasoner accepts. Data
+    properties stay on the Restrictions page until this flow can offer a
+    datatype for them.
     """
     by_id = {_uid(c["uri"]): c for c in classes}
     subject = by_id.get(st.session_state.get("_viz_crel_subject"))
     if subject is None:
         _panel_close_add()
         st.rerun()
-    if not properties:
+    if not object_props:
         st.markdown("**New restriction**")
-        st.info("Add a property first: a restriction is always on one.")
+        st.info(
+            "Add an object property first. Restrictions built here point at a "
+            "class, which a data property cannot do."
+        )
         if st.button("Cancel", key="panel_add_rest_nocancel", use_container_width=True):
             _panel_close_add()
             st.rerun()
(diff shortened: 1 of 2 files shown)


ralfbecher:
Both confirmed and fixed in e0d2e66. I reproduced each before changing anything.

**P1, the edit form.** Correct, and worse than the add-form version I fixed: this one damages data that was already right. Reproduced on the same update path:

```
BEFORE on_class: Role | on_class_uri: http://ex.org/o#Role
AFTER  on_class: None | on_class_uri: None
onClass still in graph: False
```

Re-saving an untouched, valid `qualifiedCardinality` row stripped its `owl:onClass`. Now uses `restriction_takes_on_class(new_type)`.

Rather than only swapping the call, I added an **AST test that fails on any case-sensitive `"Qualified"` membership test anywhere in `app.py`**. A plain string search was no good, since the explanatory docstring quotes the bad pattern; the AST version matches code only. That is what stops the two forms drifting apart again, which is how this one survived the first pass.

Also added an engine-level regression test that re-saves a qualified row unchanged and asserts `owl:onClass` survives.

**P2, data properties in the graph flow.** Also correct. The form was offering `foundedYear`, `hasEmail` and `hasName`, and picking one wrote `owl:someValuesFrom :Role` against an `owl:DatatypeProperty`, which I reproduced directly.

The flow now takes `object_props` only. The parameter is renamed so the call site reads as the constraint rather than relying on a comment, the docstring records why, and the empty state now says "Add an object property first. Restrictions built here point at a class, which a data property cannot do." Data properties stay on the Restrictions page until this flow can offer a datatype for them, which is the data-range handling you mention.

**Verified**

* Both new tests fail with the fixes reverted, so they pin the real defects rather than passing vacuously.
* Live: the On Property dropdown now lists only `hasDepartment`, `hasRole`, `manages`, `worksFor`. The three data properties are gone.
* 824 pass, ruff 0.16.1 and mypy clean.

Thanks, these were both real. The P1 one in particular I should have caught when I touched the add form, since it is the same predicate two hundred lines away.

PR #241, opened 2026-08-07
feat: edit annotations from the Visualization graph (#223)

### LINKED ISSUE #223: [follow-up] edit annotations from visualisation
The idea came from https://github.com/ralforion/orionbelt-ontology-builder/issues/221#issuecomment-5163324463, as I was thinking it may be better to first allow editing annotations (bringing their feature-set on-par with classes, relations and restrictions) before including them in `add entities from visualisation`. Perhaps that may help ease the code reuse process and design consistency as well.

ralfbecher:
Closes #223.

Annotations were the one thing on the graph you could not edit, and the reason turned out to be structural rather than a decision anyone made: **annotation values were drawn as anonymous nodes.** Unlike classes, individuals, data properties and SKOS concepts, they carried no type and no identity, so clicking one gave the details panel nothing to resolve and it fell through to printing the tooltip text.

## Giving an annotation an identity

Their node ids were also *counted in iteration order* (`ann_1`, `ann_2`, …), so a node was renamed whenever anything before it changed.

An annotation has no URI of its own, it **is** a triple. So it is now named by its parts, exactly as the relations and restrictions drawn as edges already are: subject, predicate, value, and the language or datatype that tells two otherwise identical literals apart. That is the same key `delete_annotation` matches on, and being content-derived it survives a rebuild.

The panel then resolves the node and offers predicate, value and language, plus delete, alongside the editors for the other kinds. This is the plumbing you predicted in the issue, and it is the same plumbing "add an annotation from the graph" will need in #221.

## An engine gap this exposed

`add_annotation` could not write a typed literal: it only ever produced `Literal(value)` or `Literal(value, lang=...)`. Since an edit is a **delete plus an add**, rewriting a value would have silently dropped a `^^xsd:date` the user never touched.

It now takes an optional `datatype`, resolved exactly as `delete_annotation` already resolved one, so a typed annotation round-trips. A language tag still wins over a datatype, since RDF literals carry one or the other, never both. The panel shows `Datatype: date (kept)` rather than leaving that invisible.

## Failing halfway

Delete-then-add can fail in the middle: `add` rejects an unusable predicate (an unbound prefix above all), and by then the original is gone. A rejected add **puts the original back** and reports the reason, so an edit that cannot be applied does not cost you the annotation.

`_apply_annotation_edit` is in the checkpoint test's `CHECKPOINTED_BY_CALLER` set for the documented reason: the panel checkpoints on the True return, and the rollback path returns False having already restored the graph.

## Verified live

Imported a class with `dcterms:source "Wikidata Q5"` and `dcterms:created "2024-01-01"^^xsd:date`:

| step | result |
| --- | --- |
| inspect the graph payload | both annotation nodes now carry `ntype="Annotation"` and a full identity including the `date` datatype |
| click the typed one | panel shows the editor: On Person, Predicate, Value, Language, `Datatype: date (kept)`, Save, Delete |
| change the value, Save | Source shows `dcterms:created "2025-06-15"^^xsd:date`, datatype intact |

843 tests pass, ruff 0.16.1 and mypy clean. `tests/test_viz_annotation_edit.py` covers the identity (content-derived not positional, two values of one predicate, language as part of the key) and the rewrite (datatype and language survive, no-op reports no change, a rejected predicate restores the original without duplicating it).

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -1734,7 +1734,11 @@ def annotation_ename(subject_uri: str, ann: dict) -> str:
         ann.get("predicate_uri") or ann["predicate"],
         ann["value"],
         ann.get("language") or "",
-        ann.get("datatype") or "",
+        # The full datatype URI, not the local name: two datatypes in different
+        # namespaces share a local name, and a non-XSD one does not resolve back
+        # from it. "uri" marks a resource-valued annotation, which is a different
+        # triple from a literal that happens to spell the same IRI.
+        "uri" if ann.get("is_uri") else (ann.get("datatype_uri") or ""),
     )
 
 
@@ -2083,9 +2087,14 @@ def _apply_annotation_edit(ont, subject_uri, ann, new_pred, new_value, new_lang)
     The datatype rides along unchanged. Without it the re-add would store a
     plain string and quietly drop a ``^^xsd:date`` that the user never touched.
     """
-    datatype = ann.get("datatype")
+    # The full datatype URI, never the display local name: a non-XSD local name
+    # does not resolve back, so deleting by it matches nothing and re-adding
+    # mints a relative ``^^<Thing>`` beside the untouched original.
+    datatype = ann.get("datatype_uri") or ann.get("datatype")
+    is_uri = bool(ann.get("is_uri"))
+    old_pred = ann.get("predicate_uri") or ann["predicate"]
     if (
-        new_pred == (ann.get("predicate_uri") or ann["predicate"])
+        new_pred == old_pred
         and new_value == ann["value"]
         and (new_lang or None) == (ann.get("language") or None)
     ):
@@ -2093,26 +2102,29 @@ def _apply_annotation_edit(ont, subject_uri, ann, new_pred, new_value, new_lang)
 
     ont.delete_annotation(
         subject_uri,
-        ann.get("predicate_uri") or ann["predicate"],
+        old_pred,
         ann["value"],
         lang=ann.get("language"),
         datatype=datatype,
+        value_is_uri=is_uri,
     )
     try:
         ont.add_annotation(
             subject_uri,
             new_pred,
             new_value,
-            lang=new_lang or None,
-            datatype=None if new_lang else datatype,
+            lang=None if is_uri else (new_lang or None),
+            datatype=None if (is_uri or new_lang) else datatype,
+            value_is_uri=is_uri,
         )
     except Exception as e:  # noqa: BLE001 - a rejected predicate must not eat the annotation
         ont.add_annotation(
             subject_uri,
-            ann.get("predicate_uri") or ann["predicate"],
+            old_pred,
             ann["value"],
             lang=ann.get("language"),
             datatype=datatype,
+            value_is_uri=is_uri,
         )
         show_message(f"Annotation unchanged: {e!s}", "error")
         return False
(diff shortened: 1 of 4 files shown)


ralfbecher:
All three confirmed and fixed in 9a41015. I reproduced each before touching anything.

**P2a, custom datatypes.** Exactly as described:

```
before: v  ^^http://types.example/a#Thing
after:  v  ^^http://types.example/a#Thing   <- original, untouched
        v2 ^^Thing                          <- new, relative, broken
```

`get_annotations` now exposes `datatype_uri` beside the local name, and everything that *acts* rather than *displays* uses the URI. The local name stays for the panel caption only.

**P2b, resource-valued annotations.** Also exactly as described:

```
before: URIRef  http://docs.example/person
after:  URIRef  http://docs.example/person  <- original survives
        Literal http://docs.example/other   <- replacement, as a string
```

`delete_annotation` and `add_annotation` both take `value_is_uri`, `get_annotations` flags it, and the panel drops the Language and Datatype fields for one, since a resource has neither. The delete path is fixed too, and it no longer reports success having removed nothing.

Worth noting the **Annotations page had both defects already** and is fixed by the same change: its delete button passed the display local name and treated an IRI object as a literal. That predates the graph editor, so this PR is now removing a bug it inherited rather than only one it introduced.

The identity carries both facts, so a resource and a literal spelling the same IRI are distinct annotations and a click cannot resolve to the wrong one.

**P3, cache version.** Fair catch, and the consequence was exactly as you say: a live session would keep serving pre-#223 payloads and clicks would fall through until something unrelated evicted them. `_graph_ver` is now 19, with a comment saying why.

**Two things my own change then broke, caught by the repo's guards rather than by me:**

* the checkpoint test flagged `add_annotation` again, because I had extracted the rollback into a nested `_restore()`, which the scope-aware check correctly treats as its own function. Inlined it rather than widening the exemption.
* mypy rejected `is_uri` because `get_annotations` was annotated `list[dict[str, str]]`. The dict genuinely holds a bool now, so the annotation says so.

**Verification**

* 6 new tests, all of which fail with the fixes reverted: full datatype URI exposed, custom-datatype edit leaves exactly one triple, resource flagged, resource edit stays a resource, resource delete actually removes, resource and literal of the same IRI are distinct.
* 849 pass, ruff 0.16.1 and mypy clean.
* Live, the graph payload now carries `http://www.w3.org/2001/XMLSchema#date` where it used to carry `date`, so the URI reaches the browser end to end.

One gap in what I checked live: I did not exercise the URI-valued panel form in the browser. Resource-valued annotations can only enter through import, and my paste-import did not take on the retry, so that path is covered by the unit tests and the engine-level repro rather than by a click. Flagging it rather than implying I clicked it.

ralfbecher:
Both confirmed and fixed in c8ade3e. Reproduced each first.

**P2, unstorable IRIs.** The worst of the set so far, because the damage outlives the edit:

```
edit returned: True
serialize RAISES: "not a uri with spaces" does not look like a valid URI, I cannot serialize this
```

One bad edit made **every export of the whole ontology** fail, with nothing pointing back at the annotation that did it.

I took your second option: `add_annotation` now refuses an unusable value before storing it, rather than the panel checking first. That covers every caller rather than the one path I happened to add, and the rewrite's existing rollback then restores the original on its own with no new machinery. The check is a new `invalid_uri_reason`, reusing the character set `invalid_name_reason` already rejects, so "what counts as unusable" is defined once.

After: the edit returns False, the original IRI is untouched, and the ontology still serializes.

**P3, bulk delete.** Confirmed: `{'applied': 1, 'errors': []}` with the triple still in the graph.

Rather than plumb `is_uri` through the bulk table, I widened the fallback in `delete_annotation`. That branch runs when the caller said *nothing* about the object, and it matched only literals — so a caller that does not know the difference silently deleted nothing. Matching a resource too is what "nothing said about the object" ought to mean, and `value_is_uri` still selects one exactly when the caller does know. The bulk table needs no new column, and the same fix covers any other caller with the same blind spot.

**Verification**

* 10 new tests, all failing with the fixes reverted: refused edit leaves the original, ontology still serializes afterwards, six unusable forms rejected (spaces, `<>`, `"`, `{}`, empty, whitespace-only), three usable ones still accepted (`http:`, `urn:`, `mailto:`), untyped delete matches a resource, bulk delete removes it, and a literal delete is unaffected by the widened match.
* 863 pass, ruff 0.16.1 and mypy clean.

**One thing I have not changed, so it is not mistaken for covered:** bulk *add* of a resource value still writes a literal, since the table has no way to say otherwise. That is pre-existing and outside what you flagged; happy to take it separately if you want resource-valued rows creatable there.

PR #242, opened 2026-08-07
fix: never let the render cap drop the entity you searched for (#234)

### LINKED ISSUE #234: [bug] some classes don't appear when searched for in `Find and centre`
I have a feeling this may be due to similar reason as https://github.com/ralforion/orionbelt-ontology-builder/issues/216 - in any case non-critical, as adding the failing item inside `Filter Nodes` 'unblocks' `Find and centre` for other items.

ralfbecher:
Closes #234. @SuperCowProducts confirmed it is the second of the two cases I asked about, and the detail they added is what pins the mechanism down rather than merely fitting it.

### What was happening

Each build loop stops at a fixed node budget, in list order, so an entity far enough down was never drawn. The Find dropdown is built from every entity in the ontology, independent of what was drawn — so it lists the class, you can pick it, and there is no node to centre on.

**Why it felt like a silent failure rather than a missing node.** The viewer only pins the camera when the picked node is present in the payload:

```js
var _pin = (!_find.released && args.focus_node && _findIds[args.focus_node]) ? args.focus_node : null;
...
options.physics.stabilization = { ..., fit: !_pin };
```

With the node absent the pin is dropped, and dropping the pin re-enables the post-layout auto-fit, which reframes the whole graph. So the movement they reported is not the app reaching for the class and failing partway: it is the graph refitting *because* the class was not there. Had it merely been hidden, nothing would have moved at all.

It also explains their Filter Nodes workaround: it is not the un-hiding that helps, it is that fewer nodes means the builder reaches theirs before the cap stops it.

### The fix

Build the picked entity first in its own loop, and let its block run even when the budget is already spent — the kinds after classes are otherwise skipped outright, so ordering alone would not reach an individual or a SKOS concept.

**Reordering rather than exempting keeps the cap honest.** The same number of nodes is drawn; the one the user explicitly named is simply not the one sacrificed to the budget. Same shape as #216, which fixed this for focus mode and left Find with the same reach problem.

### Tests

Extended the existing `test_viz_node_cap.py` harness, which already renders the real `render_visualization` and inspects the cached payload:

* a class past the cap **is** drawn when it is the Find target (fails without the fix)
* the cap still holds: the payload is still within `GRAPH_MAX_NODES`, so it is not an extra node bolted on
* a class below the cap is unaffected
* **without** a Find target that same class is still dropped, so the cap is unchanged where it protects the browser

833 pass, ruff 0.16.1 and mypy clean.

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -2975,6 +2975,28 @@ GRAPH_MAX_NODES = 500
 FOCUS_BUILD_MAX_NODES = 5000
 
 
+def prioritise_find_target(items, node_id_of, find_id):
+    """Build the entity picked in Find first, so the cap cannot drop it (#234).
+
+    Each loop stops at a fixed node budget, in list order, so an entity far
+    enough down was never drawn. The Find dropdown lists every entity regardless
+    of what was drawn, so picking one of those left the viewer nothing to centre
+    on. It then dropped the camera pin, and dropping the pin re-enables the
+    post-layout auto-fit, so the graph visibly reframed while the entity the user
+    asked for was missing: movement, and no result, which is exactly what the
+    issue describes.
+
+    Reordering rather than exempting keeps the cap honest — the same number of
+    nodes is drawn, and the one the user named is simply not the one sacrificed.
+    """
+    if not find_id:
+        return items
+    for i, item in enumerate(items):
+        if node_id_of(item) == find_id:
+            return [item, *items[:i], *items[i + 1 :]]
+    return items
+
+
 def graph_node_cap(focus_pruning: bool) -> int:
     """How many nodes a graph build may assemble (issue #216).
 
@@ -8776,6 +8798,17 @@ def render_visualization():
             focus_pruning = bool(focus_mode and focus_seed_ids)
             max_nodes = graph_node_cap(focus_pruning)
             node_count = 0
+
+            def _find_kind_is(prefix: str, _fid=_find_id) -> bool:
+                """Is the Find target one of *this* kind of node? (issue #234)
+
+                Ordering the target first only helps if its loop runs at all, and
+                the kinds after classes are skipped outright once the budget is
+                spent. Node ids carry their kind as a prefix, so the guard can
+                let a block through for the one entity the user asked for.
+                """
+                return bool(_fid) and _fid.startswith(prefix)
+
             # Why the graph is smaller than the ontology, shown above it. Empty
             # when nothing was left out.
             graph_notice = ""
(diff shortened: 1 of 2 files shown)


ralfbecher:
Confirmed and fixed in 7782d2d. You are right, and this is the half of the problem I reasoned about and then only half-solved: I relaxed the outer guard so those blocks would run, and left the inner `break`, which fires on the very first item — the prioritised one — before it can be added.

Reproduced as three failing tests against the reviewed head before changing anything:

```
FAILED ...[dprop-Data Property: findMe]
FAILED ...[ind-Individual: findMe]
FAILED ...[skos-Concept: findMe]
```

Each of the three loops now lets the target itself past its own cap check. Ordering still does the work for classes, which are built first while the budget is unspent.

**On your "displace an earlier node or otherwise add before checking the budget":** I took the second. Displacing would mean choosing a class to drop on the user's behalf, and the loops that would have to cooperate run earlier and know nothing about it. Adding past the cap costs **at most one node**, because only the single entity the user named is let through, and there is a test pinning that bound. One node over is immaterial to the browser the cap protects; a graph silently missing what you asked for is not. I have corrected the docstring, which claimed the node count was unchanged — that is true for classes only.

**One thing worth recording, because it made my first attempt at these tests wrong.** Data properties and individuals are **off by default**, and Find only lists the types that are shown. So with the toggles off the entity is not in the dropdown at all and nothing can be found — my initial test was asserting an impossible state and failing for that reason rather than the real one. The harness now enables the toggle alongside seeding the entity, which is what your repro must have been doing. SKOS is on by default, which is why that one behaved differently from the other two at first.

839 pass, ruff 0.16.1 and mypy clean.

ralfbecher:
Both confirmed and fixed in 65e4726. The first one is the important one: without it the entire change was **inert on the path a user actually takes**, and my tests were the reason I could not see that.

**P2, the cache key.** Exactly as you describe. The page builds and caches a graph on arrival with no target; picking one then changes nothing `graph_key` can see, so `needs_rebuild` stays false and the cached payload still lacks the entity. `_find_id` is now part of the key.

Your point about the tests is the sharper half of the finding. Seeding `viz_find_entity` before the first build is the one order in which the bug cannot appear, so every test I wrote was structurally incapable of catching it. I tried to fix that by driving the real two-phase path, and hit a wall worth recording: **AppTest cannot run this page twice.** Serialising widget states between runs dies in the filter multiselects:

```
ValueError: content: "I"\n is not in list
  streamlit/testing/v1/element_tree.py:730 in indices
```

That is the same family as the `segmented_control` mis-serialization already documented in `render_add_restriction`, and it is why every test in this file does a single render. So the interaction is pinned at the source instead: the assertion reads the `graph_key` line and fails if the Find target is not in it. Weaker than driving it, and I would rather say so than dress it up.

**P2, data property with a capped-out domain.** Also confirmed with a cold build. The "domain isn't displayed" guard is right in general and wrong for the target itself, whose domain is precisely the kind of class the cap drops, so it reinstated the silent no-result the prioritising exists to prevent. I took your first option: it is drawn standalone, without the domain edge it cannot have. Prioritising the domain too would pull in an arbitrary extra class and still fail when the domain is filtered out rather than capped.

**Verification**

* Both new tests fail on the previous head and pass now.
* 841 pass, ruff 0.16.1 and mypy clean.

Four rounds on this one, and each round found something the previous fix had left standing. Thanks for staying on it.

PR #247, opened 2026-08-08
fix: draw the Find target even when a later filter change hid it (#234)

### LINKED ISSUE #234: [bug] some classes don't appear when searched for in `Find and centre`

ralfbecher:
Closes #234, properly this time. @SuperCowProducts supplied their ontology and their exact class list, which is what made this findable.

### Why the previous fix was not enough

#242 fixed the render cap, and their ontology genuinely has 629 classes, so the cap was in play. But with their 25 classes selected the graph draws only **47 nodes**, nowhere near the cap. What they were still hitting is a different mechanism.

### Root cause

The picked entity was un-hidden by **rewriting the node filter, once per pick**, guarded by the find sequence. But whether that entity is visible depends on the **filter**, which changes independently and afterwards, while the sequence only moves when the **pick** changes.

Instrumenting the live interaction showed it exactly:

```
DBG find choice='Class: sin' cur_seq=1 revealed_seq=None      <- picked while visible
DBG reveal kind=class name='sin' uri='...#sin' in_sel=True    <- nothing to reveal
DBG find choice='Class: sin' cur_seq=1 revealed_seq=1         <- filter narrowed, sin now hidden
DBG find choice='Class: sin' cur_seq=1 revealed_seq=1         <- never runs again
```

So the failing order is: pick an entity while it is visible, so nothing needs revealing and the sequence is marked done; then narrow the filter so it is hidden. The pick has not changed, the sequence has not moved, and the reveal never runs again. The graph never gets the node, while the app goes on telling the viewer to centre on it, which throws `Node: 5ec7fc41cbb4 cannot be found` in the console and refits the whole graph: **movement, with nothing in view**.

Both of their observations follow from this:

* **"It works after clicking Clear value"** — clearing and re-picking is two selection changes, so the sequence moves and the reveal runs.
* **"more classes makes it likelier"** — more filter churn after a pick means more chances for the picked entity to end up outside it.

### The fix

Draw the target **regardless of the filter**, checked against the live filter on every render so it cannot go stale, the same way the render cap already exempts it.

That also stops the app silently rewriting the user's filter behind their back. Nothing else is exempt: the filter is otherwise honoured exactly as before, and there is a test asserting only the picked entity gets through.

### Verified against the reporter's own data

Their ontology, their 25-class list, picking `Class: sin` through the real widget:

| | before | after |
| --- | --- | --- |
| nodes drawn | 47 | **48** |
| `sin` in the graph | no | **yes** |
| selected / camera | none | **selected, scale 1.1** |
| console | `Node ... cannot be found` | clean |

### Tests

877 pass, ruff 0.16.1 and mypy clean. Two new tests, both failing without the fix: the target hidden by a later filter change is drawn, and the filter still hides everything else.

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -9054,7 +9054,21 @@ def render_visualization():
                 ):
                     if node_count >= max_nodes:
                         break
-                    if cls["uri"] not in visible_class_uris:
+                    # The entity picked in Find is drawn even when the node
+                    # filter hides it: you asked to see it, and a graph that
+                    # silently omits it looks broken rather than filtered.
+                    #
+                    # Checked here, against the live filter, rather than by
+                    # un-hiding it in the filter when the pick happens. That
+                    # un-hiding ran once per pick, so an entity that was visible
+                    # when picked and hidden by a *later* filter change was never
+                    # revealed again: the pick had not changed, so nothing
+                    # re-ran, and the app went on telling the viewer to centre on
+                    # a node it had not sent (issue #234).
+                    if (
+                        cls["uri"] not in visible_class_uris
+                        and _uid(cls["uri"]) != _find_id
+                    ):
                         continue
                     cls_node_id = _uid(cls["uri"])
                     disp_cls_name = _disambiguated_name(cls, cls_collisions)
@@ -9257,7 +9271,11 @@ def render_visualization():
                     # Individuals filter (issue #196). Focus mode builds the full
                     # graph and prunes afterwards, so it ignores the filter the
                     # same way the class one does.
-                    if not focus_mode and ind["uri"] not in selected_ind_uris:
+                    if (
+                        not focus_mode
+                        and ind["uri"] not in selected_ind_uris
+                        and f"ind_{_uid(ind['uri'])}" != _find_id
+                    ):
                         continue
                     ind_node_id = f"ind_{_uid(ind['uri'])}"
                     disp_ind_name = _disambiguated_name(ind, ind_collisions)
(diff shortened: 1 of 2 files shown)


ralfbecher:
Both confirmed and fixed in 4fd9394, each reproduced first.

**P2, the emptied filter.** Correct. The bypass was per-node, and the loop it sits in is guarded by `selected_classes`, so emptying the filter skipped the loop entirely and the target went with everything else. Reproduced: 0 nodes, no target.

The guard now also runs when the target is a class. Worth noting why that needed a new flag rather than reusing `_find_kind_is`: class node ids carry no prefix, unlike `ind_`/`dprop_`/`skos_`, so the kind cannot be read back off the id. It is recorded when the pick resolves instead.

This is the third time in this issue that an outer guard has swallowed a per-item bypass, which is a pattern rather than three coincidences: the cap loops, then the data-property domain guard, now the filter guard. Each one is a gate that runs before the loop and knows nothing about the target.

**P3, the filter rewriting.** Also correct, and it is exactly the behaviour this PR set out to remove: I added the render-time bypass and left the old writer in place. It edits a setting the user chose, and it goes stale the moment a later filter change hides the entity again, which was the original bug. You are also right that it masked P2, since a fresh pick refilled the selection.

Removed. Only the focus-mode seeding remains in that handler, which is about the prune rather than the filter, and a test now asserts nothing in the Find handler writes the node filter.

The two together read better than either alone. On the reporter's own ontology and class list:

| | before this PR | after |
| --- | --- | --- |
| filter | 25/629, rewritten to 26 on pick | **25/629, untouched** |
| nodes | 47, target absent | **48, target present** |
| camera | refit, nothing in view | **selected, scale 1.1** |
| console | `Node ... cannot be found` | clean |

So the filter is now left exactly as the user set it, and the entity they asked for is drawn anyway.

879 pass, ruff 0.16.1 and mypy clean. Both new tests fail without the fixes.

PR #254, opened 2026-08-09
feat: move an annotation to another resource from the graph (#251)

### LINKED ISSUE #251: [idea] move annotation from one class to the other
We can do this with relations and restrictions already by clicking on the arrow: perhaps a similar feature is also possible with annotations.

ralfbecher:
Closes #251.

Relations and restrictions can already be re-pointed from the details panel. Annotations could be edited there since #241 but not moved, because the panel *printed* the resource they hang off instead of offering it. It is a picker now.

### It reuses the rewrite rather than adding a second path

Moving is the same delete-then-add the edit already performs, just landing somewhere else, so it inherits every guard that came with it:

* the **datatype** and **language** survive the move
* a **resource-valued** annotation stays a resource rather than becoming a string
* a rejected predicate **rolls back to the original subject**, not the new one
* moving to where it already is reports **no change**, so no undo checkpoint is taken for a no-op

### Two things worth calling out

**Subjects are keyed by URI and tagged with their kind.** Local names collide across namespaces, and a picker keyed by name would move an annotation onto the wrong same-named resource silently. The tag distinguishes a class from a property that reads the same.

**An unlisted subject is offered as its own option.** The picker covers classes, properties and individuals. An annotation can sit on something else — the ontology itself, a SKOS concept — and an unlisted subject would have selected the first entry, so saving an *untouched* annotation would have moved it somewhere. I caught this because a comment I had written claimed the opposite of what the code did.

### Tests

910 pass, ruff 0.16.1 and mypy clean. Eight new tests cover the move, the datatype and resource-value cases, the rollback landing back on the original subject, the no-op, the kind tagging, URI keying across namespaces, and the unlisted subject.

### Verification note

The move logic is covered by unit tests, but I did **not** confirm the panel UI live this round: browser setup for this scenario failed repeatedly and I stopped rather than keep retrying. The change to the panel is thin — one caption replaced by a selectbox in a form that already existed and was verified live in #241 — but it is untested end to end, and I would rather say so than imply otherwise.

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -2075,7 +2075,9 @@ def restriction_references_class(restriction_type: str) -> bool:
     )
 
 
-def _apply_annotation_edit(ont, subject_uri, ann, new_pred, new_value, new_lang):
+def _apply_annotation_edit(
+    ont, subject_uri, ann, new_pred, new_value, new_lang, new_subject_uri=None
+):
     """Rewrite one annotation. Returns True when the graph changed (issue #223).
 
     An annotation is a triple, and the engine has no update for one, so an edit
@@ -2093,10 +2095,14 @@ def _apply_annotation_edit(ont, subject_uri, ann, new_pred, new_value, new_lang)
     datatype = ann.get("datatype_uri") or ann.get("datatype")
     is_uri = bool(ann.get("is_uri"))
     old_pred = ann.get("predicate_uri") or ann["predicate"]
+    # Moving an annotation to another resource is the same delete-then-add, just
+    # landing somewhere else, so it rides the same guards and rollback (#251).
+    target_uri = new_subject_uri or subject_uri
     if (
         new_pred == old_pred
         and new_value == ann["value"]
         and (new_lang or None) == (ann.get("language") or None)
+        and target_uri == subject_uri
     ):
         return False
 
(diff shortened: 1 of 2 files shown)


ralfbecher:
Confirmed live, so the verification caveat above no longer applies.

Drove the whole flow in a browser: applied the Organization template, added `dcterms:contributor "Wikidata Q5"` to **Department** from the Annotations page, then in the Visualization panel clicked the annotation node and moved it.

| step | result |
| --- | --- |
| click the annotation node | panel shows the editor with **On → `Department [Class]`** above Predicate / Value / Language |
| open the picker | all ten resources offered, each tagged with its kind |
| choose `Person [Class]`, Save | the node's edge becomes **`Person -> Wikidata Q5`**, and its identity updates to `...#Person` |
| Source | `:Department` no longer carries it; `:Person` has `dcterms:contributor "Wikidata Q5"` |

The panel then reads "This annotation was edited or removed. Click a node to pick one up.", which is right and matches what the relation and restriction editors do after a change: the selection still points at the old identity, and that identity no longer names anything.

Two incidental notes from driving it:

* the details panel starts collapsed, which is what defeated my earlier attempts at this, not the feature
* `_pad_option` pads option labels, so an anchored text match on a dropdown entry never matches. Worth knowing for anyone automating these pickers later

PR #255, opened 2026-08-09
feat: pick the class for a restriction value instead of typing it (#250)

### LINKED ISSUE #250: [idea] class selection menu under `Value` when editing restrictions in visualisation
This is mainly because
1. The selection menu is very convenient
2. The selection menu is everywhere else

Otherwise, manually typing the full class name without any autocomplete works.

ralfbecher:
Closes #250.

`someValuesFrom` and `allValuesFrom` point at a class, but the value was free text while everywhere else offered a picker, so the class name had to be typed in full with no autocomplete and nothing to catch a typo.

### Why it was free text

The old comment on the field said it outright: *"a form cannot swap the widget when the type picker changes"*. A form batches until submit and does not rerun when a widget inside it changes, so the value field could not follow the type and had to stay free text for every type.

The type picker now sits **outside** the form, so changing it reruns and the right widget appears:

| type | value field |
| --- | --- |
| `someValuesFrom`, `allValuesFrom` | class picker |
| the cardinalities | free text |
| `hasValue` | free text |

Cardinalities and `hasValue` are a number and a literal, and were never the complaint, so they are untouched.

A value the class list does not hold — an imported class, an external URI — is offered as itself, so an unchanged row still round-trips.

### Verified live

* the edit form shows **Value (Class)** as a dropdown with no free-text Value field
* switching the type to `exactCardinality` swaps it to free text, and back to `someValuesFrom` swaps it to the picker: the thing the old comment said could not be done
* all three dropdowns filter as you type: `Per` narrows to `Person`, `o` narrows the property list to three

### Tests

911 pass, ruff 0.16.1 and mypy clean.

The existing tests drove a text input that class-valued rows no longer have, so they now drive the picker and let the extra rerun happen when they change the type. The empty-value guard moves to `hasValue`, where free text remains, since a picker cannot submit an empty class — with a new test asserting exactly that.

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -5394,10 +5394,13 @@ def render_restriction_form(ont, rest, form_key, classes, properties, on_close=N
     )
     on_options, on_lookup = _slot_options(classes, rest.get("on_class_uri"))
 
-    # The value stays free text: it can be a class, a cardinality or a literal,
-    # and a form cannot swap the widget when the type picker changes. It is
-    # pre-filled as a full URI in the two cases where a bare local name would
-    # not round-trip (review P2):
+    # The value is a class, a cardinality or a literal depending on the type, so
+    # the type picker sits *outside* the form below: a form batches until submit
+    # and does not rerun when a widget inside it changes, so the value field
+    # could not follow the type and had to stay free text for every type
+    # (issue #250). Outside, changing the type reruns and the right widget
+    # appears. It is pre-filled as a full URI in the two cases where a bare local
+    # name would not round-trip (review P2):
     #   * an imported entity, since a local name resolves into the base
     #     namespace and would point at a different thing;
     #   * any resource-valued hasValue, since add_restriction stores a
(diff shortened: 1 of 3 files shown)


ralfbecher:
Confirmed and fixed in 9f24149. Reproduced it first, exactly as you describe: opening the `hasValue :alice` row, switching to `someValuesFrom`, and the picker offered `Person` **and** `http://test.org/ont#alice`, with Save writing `owl:someValuesFrom :alice`.

The picker was carrying the row's own value URI in unconditionally. That is right when the value already names a class — an imported class or an external URI has to survive an unchanged save, which is what `_slot_options` is for — and wrong for every other type, because a `hasValue` value is an individual or a literal.

It is now carried in only when the existing type is already class-valued. For anything else the picker is the class list alone.

After the fix, driving the same path:

```
picker options: ['Person']
saved -> someValuesFrom value: http://test.org/ont#Person
```

912 pass, ruff 0.16.1 and mypy clean. The new test opens that hasValue row, switches the type, and asserts the picker offers only `Person`; it fails against the previous commit.

Good catch — the round-tripping case and the type-change case pull in opposite directions here, and I had only been thinking about the first.

Issue #152, opened 2026-07-19
[idea] edit relations and restrictions from the Visualization graph (stage 2)


SuperCowProducts:
Stage 2 of the original request. Stage 1, the on-page editors for Relations and Restrictions, shipped in v1.17.4 via #188.

## Remaining scope

Clicking a relation or restriction in the Visualization graph should lead to its editor, either through the `Open full editor` button or by editing it directly in the Details panel, the way classes and individuals already work.

## Why it does not work today

- `_type_to_page` in `orionbelt_ontology_builder/app.py` maps only Class, Object Property, Data Property, Individual and SKOS Concept, and `show_view` gates the button on membership in that map.
- `_render_panel_entity_editor()` resolves those same five kinds. Structural edges fall through to the read-only tooltip text.
- Selecting an object property edge opens an editor, but it edits the property definition rather than the relation triple that was clicked, so it does not satisfy the request.

## Suggested first step

Not every restriction is drawn in the graph, but custom restrictions do appear as dashed edges. Giving those the panel editor is a useful increment and avoids blocking on the larger question of representing every restriction as its own graph element.

## Original request

From @SuperCowProducts:

> I have found the editable `Details` sidebar for classes and individuals very useful, so I thought allowing editing of relations and restrictions right from Visualization as well would be handy too.

> For there to be an `Open full editor` button available when clicking on restrictions or relations, each one of these entities would indeed need its own on-page editor. Also, good point not all restrictions may be visible on Visualization. I am using custom restrictions which do show up for me (as dotted arrows in blue): if it's feasible to provide an editing menu even just for those, I think that would be a great start.

ralfbecher:
Thanks for the idea! Before building this I'd like to make sure I target the right thing, because there are two quite different ways to read it.

Some context: today the Relations and Restrictions pages only let you **view and delete** items. Unlike Classes, Properties, and Individuals, they have no per-item **Edit/Delete** flow, and the "Open full editor" button in the Visualization side panel doesn't appear for them.

So did you mean:

1. **An on-page editor for relations and restrictions** — the ability to pick a specific relation/restriction, see its details, and edit or delete it right on the Relations/Restrictions page (bringing them up to parity with how classes/properties/individuals work). This works whether or not you're using the graph, and pairs nicely with the new search from #148.

[...]

PR #261, opened 2026-08-09
feat: delete entities from the Visualization panel

### LINKED ISSUE #222: [idea] delete entities from visualization
The parallel to https://github.com/ralforion/orionbelt-ontology-builder/issues/221, would be a lot less work to implement.

ralfbecher:
Closes #222, the counterpart to #221: what the details panel can select,
it can now delete, without going to find it again on its page.

## What can go

| Selected | Deletes |
| --- | --- |
| Class, Object Property, Data Property, Individual | the entity, by URI |
| Class Relation | the triple |
| Restriction | the axiom, matched by property, type and value |
| Annotation | already had a Delete, from #223 |

## Two clicks, and the same confirmation the pages use

From the graph you are one click away from an entity you were only
looking at, and `delete_class` takes its references with it, so the
second click is preceded by the impact summary `confirm_delete` builds.
The button sits outside the editor's form: a form batches until submit,
so a confirmation drawn inside one could not appear until something else
was submitted.

An axiom has no URI for `get_delete_impact` to look up, so those confirm
with what is going instead, and say whether anything actually went.
`delete_restriction` answers that itself; `remove_class_relation` returns
nothing, so the graph is measured around the call rather than assumed. An
edge can outlive the axiom behind it, and a delete that removed nothing
while reporting success is the failure #152 fixed at the page level.

## The selection that outlives the entity

The graph component keeps reporting the deleted node as selected, having
not been told otherwise, so clearing the stored selection alone lasted
exactly one rerun and the panel reopened on the entity that had just
gone. The node id is remembered and read as no selection until a
different node arrives.

## Verified in the browser

Selected a class: the summary named the 11 triples, 2 lost property
ranges and 2 annotations that would go with it. Cancel left it alone.
Added a throwaway subclass, selected it, confirmed: the entity was
removed, the graph redrew without it, and the panel reset to "Click a
node to see details".

## Tests

`tests/test_viz_delete.py` drives the panel through AppTest: the first
click only asks, confirming deletes, cancelling keeps, the same path
works for an individual and a property, the revision moves (so it is
undoable, redraws the graph and is autosaved), both edge kinds delete
while their endpoints stay, and the dropped node stops counting as
selected.

`test_mutation_checkpoints` gains the two panel edge editors, which hand
their delete to the shared helper that owns the checkpoint.

944 passed, ruff and mypy clean.

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -9585,6 +9585,10 @@ def render_visualization():
         # the loop before the per-node bypass could run (issue #234 review).
         _find_is_class = False
         focus_seed_ids: list = []
+        # Declared with the rest, not left to the branch that draws the picker:
+        # with focus mode on and every focusable type switched off, that branch
+        # does not run and the page crashed reading this further down (P1).
+        focus_seeds: list = []
         focus_depth = 0
         _find_col, _filter_col = st.columns([1, 3])
         with _find_col:
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -3136,7 +3136,7 @@ def _viz_selection_key(sel):
     return (sel.get("ntype"), sel.get("ename"), sel.get("nodeId"))
 
 
-def _viz_live_selection(live, dropped):
+def _viz_live_selection(live, dropped, revision):
     """Reconcile the component's reported selection with a deleted one (#222).
 
     Returns ``(selection, dropped)``: what the panel should show, and the marker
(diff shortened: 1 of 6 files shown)


ralfbecher:
Confirmed and fixed in 3b0c80e.

The comparison was `_live_sel.get("nodeId") == st.session_state.get("_viz_dropped_node")`, and with no delete pending both sides are `None`, so every edge click compared equal and was discarded. Worse than the delete buttons being unreachable: no edge could open the details panel at all, so relation, restriction and object-property editing from the graph was broken outright.

Selections are now identified by `(ntype, ename, nodeId)`, which nodes and edges both carry, and the marker is only consulted when one is set.

I pulled the reconciliation out into `_viz_live_selection(live, dropped)` so the path can be tested without rendering the page, which is exactly the gap you named: the existing tests drive the panel helpers directly and never cross the component-to-panel boundary. Four tests cover it now, including one that asserts an edge pick survives when nothing is pending, and one that only the deleted edge is suppressed.

Also verified in the browser: clicking a `subClassOf` edge opens the relation editor with Subject, Relation, Object, Save and Delete relation.

948 passed, ruff and mypy clean.

ralfbecher:
Confirmed and fixed in f4d8001.

`focus_seeds` is bound only inside `if focus_mode and focus_targets:`, so the `elif focus_mode:` branch — focus on, every focusable type off — skipped it and my hidden-note code read it below. Both are persisted settings, so a session can genuinely come back in that state; it is not a contrived combination.

Declared it alongside `focus_seed_ids` and `focus_depth`, which are the other focus variables the picker branch sets.

`tests/test_viz_focus_no_targets.py` renders the page in exactly that state. I checked it bites: removing the declaration again gives `UnboundLocalError: cannot access local variable 'focus_seeds' where it is not associated with a value`, which is your traceback. A second test pins the "Enable Classes, Individuals or SKOS above to pick focus nodes" message, since an empty graph with no explanation reads as the focus itself being broken — not crashing is the floor here, not the goal.

955 passed, ruff and mypy clean.

PR #262, opened 2026-08-09
refactor: split app.py into ui, pages and the shell

ralfbecher:
`app.py` had reached 11.5k lines with three layers tangled together. It is
now **650 lines**.

| | Lines |
| --- | --- |
| `app.py` (page config, CSS, sidebar, `main()`) | 650 |
| `ui.py` (shared helpers) | 4692 |
| `pages/visualization.py` | 2063 |
| `pages/properties.py` | 886 |
| `pages/import_export.py` | 679 |
| `pages/skos.py` | 608 |
| the remaining eight pages | 14 to 497 each |

The layering is the point, not the line count: `ui` knows nothing about
the pages, no page imports another, and `app.py` wires them together. A
page can no longer quietly reach into another page's internals, because
there is nothing to reach through.

`app.py` re-exports everything, so the top-level shims and the tests
still address `app.<name>`.

## What this turned up

**A broken graph that 969 tests did not notice.** The component directory
was resolved from `__file__`, which had become `pages/lib/graph_viewer`.
Nothing imports a custom component, and the path is only built at render
time, so the whole suite passed while the Visualization page showed an
error. The package now exposes `PKG_DIR` and the assets hang off that.
`tests/test_package_assets.py` asserts the bundled files resolve and that
no UI module builds a path from its own `__file__`; it fails with the old
line restored.

**Monkeypatching became ambiguous.** `from .ui import X` binds a *copy*,
so patching one module leaves the others' calls untouched — and these
names are read from several. Tests go through a `patch_ui` fixture that
sets the attribute on every UI module, which is what `setattr(app, ...)`
meant when there was one.

**The source-scanning tests were reading the wrong file.** The AST guards
(mutation checkpoints, the two dropdown rules) and six viz tests named
`app.py` by path. Left alone, the guards would have gone on passing while
covering only the half that stayed behind. They now find the UI modules
by their streamlit import, so a page added tomorrow is covered without
anyone remembering. Confirmed by breaking a check inside `ui.py` and
watching the guard fail.

## Verified

972 passed, ruff and mypy clean. All thirteen pages render in the running
app, the graph draws its nodes, and the sidebar logo loads — the two
assets that the `__file__` change would have broken.

## Not done here

`ui.py` is still 4.7k lines and wants splitting again along its own seams
(state, option builders, the shared entity forms). That is a separate
change; this one is already wide.


ralfbecher:
Both fixed, and the P1 was worse than the line count suggests — thank you.

**P1**: six statements, all in `pages/import_export.py`, now `..ontology_manager` / `..templates`. Templates, Upper and Reference Ontologies, New Ontology and Clear Ontology would each have raised on first use.

**P3**: `test_dropdown_search_ranking` reads the UI sources through `tests/sources.py` now, like the other AST guards.

On "how do we know the rest is covered", I ran your three layers:

**1. Mechanical equivalence.** Independently confirmed against `3f8b304`: 163 top-level definitions before the split, **160 AST-identical** after, 0 missing, and all 35 module constants present. The three that differ are the intended ones — `main` (page table), `render_visualization` (asset path), `render_import_export` (this fix).

**2. Runtime smoke over hidden branches.** `tests/test_page_smoke.py` renders every page and every sub-tab: 42 cases, with the tab list read off the source so a tab added later is covered without anyone remembering. With the bad imports restored it fails on `import_export:Templates`, `:Upper Ontologies` and `:Reference Ontologies`.

That leaves the three inside *button handlers*, which no render reaches, so `test_package_assets` gained a check that every relative import in the package resolves — including the ones nested in functions. It reports `import_export.py:423 .templates` with the bug back.

**3. Source guards.** All of them read `sources.ui_sources()` now, which finds UI modules by their streamlit import, so a page added tomorrow is covered.

Both new guards were verified by reintroducing the bug rather than assumed. 1015 passed, ruff and mypy clean.

PR #284, opened 2026-08-14
fix: keep the focus on an entity you rename (#275)

### LINKED ISSUE #275: [minor] renaming the focused class whilst in `Focus on one node` removes it from view

ralfbecher:
Closes #275.

## The bug

With "Focus on one node" switched on, renaming the class you are focused on
dropped it out of the graph and the view jumped to some unrelated class.

The focus seeds are held as the labels the "Focus node(s)" multiselect shows
(`"Class: Person"`), not as identities, and a rename mints a new URI. On the next
render the old label names nothing, so the seed was pruned twice over:

* `prune_reused_focus_seeds` sees a seed whose recorded node id no longer matches
  and drops it. That guard was written for ontology swaps (#180), but a rename
  trips it identically.
* the label membership filter drops it too, and the empty result falls back to
  `focus_labels[0]`, an arbitrary first class.

That fallback was then written back over the user's selection and persisted to
the linked file's saved state, so the loss survived a restart.

## The fix

Nothing in the rebuilt state ties the old and new names together (the label
changed, and the node id is a hash of the URI, which changed with it), so the
rename itself now leaves a note behind: `viz_note_rename` records old node id to
new node id. The next Visualization render consumes it via
`follow_focus_seed_renames`, swapping the seed for the label the entity now
answers to and bringing its recorded id along so the reuse prune still
recognises it.

A note is only followed when the renamed entity is still in `focus_targets`,
which keeps two edge cases honest: an entity whose type is toggled off is left
to the existing prune, and an undone rename re-points nothing. Rename chains
resolve to the last name, and a rename back to an earlier name terminates rather
than cycling.

Covers classes, individuals, data properties and SKOS concepts, from both the
page edit forms and the graph side panel. `focus_targets` and the rename notes
now build node ids through one shared `viz_node_id`, since the two have to agree
on the id or a rename would re-point nothing.

## Verification

11 new tests in `tests/test_viz_focus_rename.py`. Full suite: 1068 passed. ruff
format and check clean, mypy clean.

Also verified in the running app rather than only in tests, since the bug lives
in the session state flow across a rerun: loaded gist (181 classes), switched on
focus mode seeded to `Class: Account`, renamed it to `ZzzAccount` so that it
sorts last, and the seed followed to `Class: ZzzAccount`. Before this change it
fell back to `Class: Address`.

## Noticed but not touched

`viz_find_entity` (the "Find and centre" pick) and `_viz_last_selection` also
keep stale labels and ids after a rename. Out of scope for the reported bug,
happy to fold them in if wanted.

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -153,6 +153,7 @@ from .ui import (  # noqa: F401
     display_flash_message,
     filter_entry_token,
     focus_seeds_from_selection,
+    follow_focus_seed_renames,
     format_label_name,
     get_ontology_manager_class,
     graph_node_cap,
@@ -195,6 +196,8 @@ from .ui import (  # noqa: F401
     viz_hidden_caption,
     viz_hidden_note_style,
     viz_mark_ontology_seen,
+    viz_node_id,
+    viz_note_rename,
     viz_ontology_was_replaced,
     viz_sync,
 )
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -154,6 +154,7 @@ from .ui import (  # noqa: F401
     filter_entry_token,
     focus_seeds_from_selection,
     follow_focus_seed_renames,
+    follow_renamed_node_ids,
     format_label_name,
     get_ontology_manager_class,
     graph_node_cap,
--- a/orionbelt_ontology_builder/ui.py
+++ b/orionbelt_ontology_builder/ui.py
@@ -778,6 +778,20 @@ def _renamed_node_id(node_id, renames):
     return node_id
 
 
+def follow_renamed_node_ids(node_ids, renames):
+    """Where ``node_ids`` ended up after the recorded renames (issue #275).
+
+    The seeds saved against a linked file are node ids rather than labels
+    (#164), and one saved before a rename made this session resolves to nothing
+    when it is read back. Same notes and same chain-following as
+    :func:`follow_focus_seed_renames`, one level down: that works in the labels
+    the seeds are shown under, this in the ids they are stored as.
+    """
+    if not renames:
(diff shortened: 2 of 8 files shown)


ralfbecher:
Both findings were real, and both are fixed in 1f35bb7.

**P2, the file-restored seeds.** Confirmed: the notes were consumed only by the
seeds already in session, and the seeds saved against a linked working file are
node ids restored further down (#164). Renaming that entity before the first
Visualization render of a session hit exactly the hole described: the note was
discarded because `_viz_cfg_focus_seeds` was absent, and then the saved id no
longer resolved. Taking the "delay the pop" option rather than a second pop, so
there is one place the notes are consumed: `_renames` is popped once and both
seed sources read it. The id path goes through a new `follow_renamed_node_ids`,
which follows the same chains as `follow_focus_seed_renames`, one level down.

**P3, notes not scoped to the ontology.** Also fixed, by clearing the notes when
the render sees a replacement, reusing the counter check the class filter
already relies on (`viz_ontology_was_replaced`, #180). Two things worth noting
about the shape of that guard:

* A file switch was already covered, since `_clear_viz_file_session_state` drops
  the seeds and the notes together.
* Undo counts as a replacement, which is the outcome we want here: an undone
  rename should leave the seed on the name that is back in force, and now it
  does so explicitly rather than by relying on the target id being absent.

The one case still outside the guard is a replacement on a session that has
never rendered Visualization, where `viz_ontology_was_replaced` returns False
because it has no previous counters to compare. That is pre-existing and applies
equally to the reuse prune itself, so the notes are no weaker than the guard
they sit in front of.

Three new tests (14 in `tests/test_viz_focus_rename.py`), full suite 1071
passed, ruff format and check clean, mypy clean. The P2 path is covered by unit
tests rather than live, since it needs a linked working file, which the browser
session does not have.

PR #310, opened 2026-08-21
feat(viz): the graph's panels float over the canvas

ralfbecher:
Everything that opened on the graph page was paid for out of the canvas: the details panel took a quarter of the row's width, the reopen toggle took a sliver of it, and the Node options picker pushed the graph down far enough to squeeze it to its 300px floor. Opening a panel to read one node cost you the view you opened it from.

All three now sit on top of the canvas as solid cards, and the canvas claims the whole row whatever is open.

## Details

**Solid, not translucent.** Chips and labels read over a busy graph are unreadable. The two card palettes follow the graph's own light/dark themes.

**The details panel** pins both its edges, so `max-height` has a length to resolve against and a long editor scrolls inside the card instead of running past the graph. It sizes to its content, so a two-line panel is a two-line card.

**Node options** hangs off the bottom edge of its own summary, so the picker still reads as part of the control that opened it. It sits above the details panel: it is the thing the click just opened. Its multiselect dropdown renders in a body-level portal, so the card's `overflow-y` does not clip it (checked in the browser).

**The reopen toggle** floats in the same corner as the panel it reopens, so the canvas does not change width when you open or close the panel.

## Testing

The CSS moved out of the render path into `graph_overlay_css()` so the hooks can be tested. Every hook is Streamlit's markup rather than this app's, and one that stops matching fails silently, so `tests/test_viz_overlays.py` pins them.

That file also back-fills a guard for the browser-storage rule from #309. A review suggested that rule's `streamlit_local_storage` substring never matches because the component is declared as `st_local_storage`. It does match: Streamlit serves a component from `/component/<module>.<name>/`, so the live URL is `/component/streamlit_local_storage.st_local_storage/index.html` and contains both names. Confirmed against the running app, where the selector matches exactly one element and its container computes to `position: absolute; height: 0`. The new test ties the substring to the module the app imports, which is the part that could actually drift.

## Verification

Driven in the running app at 1400x900: panel open and closed, Node options open over both, a node selected with the status bar and its "Open full editor" button, and fullscreen (the row is now a positioning context, but the canvas goes `position: fixed` in fullscreen and still covers the viewport).

pytest (1498 passed), ruff format/check, and mypy are clean.

### DIFF
--- a/orionbelt_ontology_builder/ui.py
+++ b/orionbelt_ontology_builder/ui.py
@@ -117,6 +117,17 @@ _CUSTOM_CSS = """
         border-radius: 4px;
         color: #721c24;
     }
+    /* CSS injected through st.markdown leaves an empty element behind, which
+       is invisible but still takes a slot in the page column's 16px gap grid.
+       Four of them ride above the graph, so the canvas started ~64px lower than
+       it had to. Nothing to show, so take them out of the flow. */
+    [data-testid="stElementContainer"]:has(
+        > [data-testid="stMarkdown"]
+        > [data-testid="stMarkdownContainer"]
+        > style:only-child
+    ) {
+        display: none !important;
+    }
     /* The browser-storage component has nothing to show, but Streamlit still
        lays its iframe out — a 26px band, plus the column's 16px gap, wherever
        the write happens to be called (on the graph page, right above the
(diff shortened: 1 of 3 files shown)


ralfbecher:
Pushed three follow-ups (fa6ab06).

**The overlays were sitting on the canvas's own toolbar.** The download and fullscreen buttons live in the same top-right corner the cards claimed, and the overlays are outside the iframe, so they won every hit test: with the panel open both buttons were dead, with it closed the reopen toggle covered fullscreen. Confirmed by hit-testing each button's centre in the running app, and confirmed reachable again afterwards in both states. The overlays now start below the toolbar, and the test derives the clearance from the buttons' own CSS so the two cannot drift.

**Opening Node options slid the graph down, then snapped it back.** Streamlit opens an expander with a height animation driven from JavaScript, aimed at the height of a body that is out of the flow here. CSS `transition: none` cannot touch a script animation and neither could zeroing `::details-content`; an `!important` declaration outranks it. Measured per frame: 73px of travel before, 0px after.

**Space above the canvas.** Every CSS-only `st.markdown` leaves an empty element that still takes a slot in the page column's 16px gap grid, and four of them ride above the graph. Out of the flow now. That pulled the first line of the page up under Streamlit's header, which is an opaque bar painted over the page rather than behind it, so the top padding grows to clear it: it was only ever wide enough because those empty elements were padding it out.

pytest (1502 passed), ruff, and mypy are clean.

PR #316, opened 2026-08-23
fix(viz): the details panel's link points at the whole value

### LINKED ISSUE #313: [bug] clickable link doesn't expand after opening in browser
The link in blue on the top right will open `https://webeep.polimi.it/cours` rather than the full link:

Image

ralfbecher:
Fixes the clickable link in the graph's Details panel (issue #313).

## The bug

An annotation node draws a label cut to 30 characters so it fits on the node, and the details panel headed with that label as plain text. Streamlit linkifies a bare URL, so a cut URL became a link to the cut URL: clicking it opened `https://webeep.polimi.it/cours` instead of the page the annotation names.

## The fix

- The node carries its whole value (`flabel`) alongside the drawn label, for annotation nodes and for literal nodes in the triple view.
- The viewer sends it with the selection, on a click and on a Find & centre alike.
- The heading is an explicit link: the text stays as short as the node drew it, the destination is the value in full. Markdown does not linkify inside a link's own text, so nothing rewrites it back. Labels that are not URLs are unlinked, and markdown syntax in a label is escaped rather than rendered.

## Verified

`pytest` (1511 passing, 7 new), ruff, mypy, plus a live run: clicking the annotation node yields one anchor with text `https://webeep.polimi.it/cours...` and href `https://webeep.polimi.it/course/view.php?id=123456`.

Closes #313

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -180,6 +180,7 @@ from .ui import (  # noqa: F401
     log_error,
     maybe_restore_autosave,
     missing_required,
+    panel_heading_markdown,
     panel_subject_uri,
     parent_option_index,
     parse_filter_text,
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -180,7 +180,7 @@ from .ui import (  # noqa: F401
     log_error,
     maybe_restore_autosave,
     missing_required,
-    panel_heading_markdown,
+    panel_heading_html,
     panel_subject_uri,
     parent_option_index,
     parse_filter_text,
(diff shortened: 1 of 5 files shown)


ralfbecher:
Good catch on the P2, and the suggested shape needed one correction: escaped *inline* HTML does not stop the linkifier. Streamlit still markdown-parses the text between inline tags, so a `<b>`-wrapped bare URL gets its own `<a>` nested inside ours, and the browser flattens that to the truncated link again. That is what the first attempt at this fix did, and it reproduced the bug.

An HTML **block** does not have that problem: markdown copies a block through raw, so nothing inside is linkified. The heading is now a single `<p><strong>...</strong></p>`, with the anchor built in code when the value is a URL and no anchor at all otherwise.

Verified in a browser, since none of this is observable from the markdown string alone:

| value | rendered |
| --- | --- |
| `https://webeep.polimi.it/course/view.php?id=123456` (cut) | one anchor, text `https://webeep.polimi.it/cours...`, href the whole URL |
| `https://webeep.polimi.it/course/view.php?id=123456 notes` (cut) | no anchor |
| `Course` | no anchor |

Tests updated: the prose case is pinned, and one test guards the property the whole thing rests on (the heading starts the line, so it is a block).

PR #320, opened 2026-08-23
feat(viz): copy what is selected, from the canvas or the bar

### LINKED ISSUE #312: [idea] make text in bottom bar that appears after clicking something in visualisation selectable
Alternatively specific parts of text that appear (e.g. annotations) could be made copiable. Handy feature now that the sidebar doesn't automatically open anymore.

ralfbecher:
Makes the selected text copiable (issue #312), on both surfaces the issue names.

## Why

Both places a selection is reported cut it to fit: a node draws a label short enough for the node, and the bar under the graph ellipsises its one line. An annotation's value is usually the thing worth copying, and it is exactly what those cuts eat — it could be read on screen but not taken away.

(The bar's text was already selectable, checked in a browser: a double-click selects a word and nothing in the CSS blocks it. What was missing is the part the ellipsis hides, and a one-click copy.)

## What

**On the canvas** — a copy button beside download and fullscreen. It appears with a selection and goes with it, and hands over the whole thing:

```
http://xmlns.com/foaf/0.1/
rdfs:isDefinedBy: http://xmlns.com/foaf/0.1/
```

the untruncated label first, then the tooltip's own lines, so the value and what it hangs on read as they do on screen. It works whichever panel is open. The async clipboard is used where there is one, and the old selection-based copy where there is not: the desktop webview has no clipboard API, and a button that silently does nothing there is worse than none. The icon becomes a tick for a moment, or says so if the copy failed.

**On the bar** — the whole line is now in a `title` tooltip, and a copy popover rides at the end of the row holding the selection as a code block, which is what carries Streamlit's own copy-to-clipboard button. A hand-written one would not work: Streamlit strips `onclick` from `st.markdown` HTML (checked). The bar also escapes its text on the way in now, which it did not before, and shows the whole label rather than the node's cut one.

Two small cleanups came with it: the three canvas buttons share one CSS rule and one theming helper instead of three copies each, and only the last button in the status-bar row is rounded, so two buttons no longer put a seam in the middle.

## Verified

In the running app: selecting an annotation shows the button, clicking it reports success from the clipboard API, and pasting into a field gives back exactly the two lines above. The bar's tooltip carries the whole line, and the popover opens with the same text and its copy icon. Screenshots of both were taken while testing.

`pytest` (1528 passing, 7 new), ruff, mypy. The existing toolbar-clearance guard in `test_viz_overlays.py` had its CSS parser updated for the shared rule, and now covers the new button too.

Closes #312

### DIFF
--- a/orionbelt_ontology_builder/lib/graph_viewer/index.html
+++ b/orionbelt_ontology_builder/lib/graph_viewer/index.html
@@ -15,24 +15,20 @@ html, body { margin: 0; padding: 0; overflow: hidden; background: #fff; }
 #legend .line { display: inline-block; width: 20px; height: 2px; vertical-align: middle; margin-right: 2px; }
 #legend .line.dashed { border-top: 2px dashed; height: 0; }
 #legend .line.solid { border-top: 2px solid; height: 0; }
-#download-btn {
-    position: absolute; top: 8px; right: 48px; z-index: 10;
+#fullscreen-btn, #download-btn, #copy-btn {
+    position: absolute; top: 8px; z-index: 10;
     width: 32px; height: 32px; border: none; border-radius: 6px;
     background: rgba(255,255,255,0.92); cursor: pointer;
     display: flex; align-items: center; justify-content: center;
     opacity: 0.7; transition: opacity 0.2s;
 }
-#download-btn:hover { opacity: 1; }
-#download-btn svg { width: 18px; height: 18px; fill: #555; }
-#fullscreen-btn {
-    position: absolute; top: 8px; right: 8px; z-index: 10;
-    width: 32px; height: 32px; border: none; border-radius: 6px;
-    background: rgba(255,255,255,0.92); cursor: pointer;
-    display: flex; align-items: center; justify-content: center;
-    opacity: 0.7; transition: opacity 0.2s;
+#fullscreen-btn { right: 8px; }
+#download-btn { right: 48px; }
+#copy-btn { right: 88px; }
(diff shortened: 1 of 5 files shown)


ralfbecher:
Both confirmed and fixed.

**P2 — the icon only overlaid the bar inside a column.** Reproduced exactly as described: with an annotation selected (no View button, so the bar has the row to itself) the iframe rendered *below* the black strip at full width — measured top 1098 against a bar of 1082–1118 — while the bar reserved 44px of padding for an icon that was not over it. The extra height also pushed the sticky bar up the page, which is the jump reported separately.

A negative margin alone did not fix it: the flow gap still counted, so the bar still moved. The bar and the button now share one container, which is the one thing the button can be positioned against in either layout, and the container joins the sticky rule list so the bar still pins to the bottom. Measured after the fix:

| | bar top | icon |
| --- | --- | --- |
| no selection | 1094 | none |
| annotation (no View button) | 1094 | 36x36 at the bar's right end, inside it |
| class (View button beside) | 1094 | same, inside it |

**P2 — an edge lost the canvas button on the next render.** Also right, and the rerun I first tried was not enough to show it: the args were unchanged so the component never re-mounted. Toggling the details panel does re-mount it, and that reproduced the loss. Python now passes the selection's copy text to the component, which uses it for anything it cannot restore by id. After a genuine re-mount with an edge selected — the instrumentation I had patched into the frame was gone, so it really was a new frame — the button is still there offering `subClassOf`.

Three tests added for these, including one that fails if the positioning is scoped to `stColumn` again. 1534 passing, ruff and mypy clean.

Issue #190, opened 2026-07-25
[bug] new entities cannot be found in Visualization until restart


SuperCowProducts:
Just a tricksy little edge-case which might have been missed when fixing https://github.com/ralforion/orionbelt-ontology-builder/issues/180.

[UPDATE] Hmm, curious to see it isn't happening any more with classes I'm creating now. Maybe just temporary glitch? Or maybe I caused an update by creating a relation 💭

ralfbecher:
Thanks for the report, and for the honest follow-up. Chasing it turned up a real bug with exactly this signature, though I can't yet say it is the one you hit.

The app keeps a revision counter that is bumped in one place, `save_checkpoint()`. Three things key off it: the undo history, the Visualization's cache key (so the canvas redraws), and autosave, which skips writing when the revision has not moved. Six actions changed the ontology without bumping it: adding and removing an import, and the Advanced page's property chain, disjoint union, AllDifferent and hasKey.

Anything added that way stayed invisible to the graph, could not be undone, and was never saved. It survived in memory only, until some later action that did checkpoint swept it up, which is exactly your "maybe I caused an update by creating a relation" observation: creating a relation checkpoints, and that one bump refreshes the graph and flushes the earlier change to storage.

Fixed in #191, along with a test that walks the app's syntax tree and fails with the file and line of any future mutation that skips the checkpoint, so this cannot come back quietly.

One caveat on the diagnosis: you mentioned classes, and creating a class does checkpoint, so your exact path may have been something else. If it happens again, the useful detail is which page you created the entity on. Leaving this open until you have had a chance to see whether it recurs on the fixed build.

SuperCowProducts:
Ok, just tested this in an empty ontology and what happens now is the newly-created class is added to the graph. I think we'd want to let the user decide whether to view newly-created classes, but other than that all good 👍 

Making the newly-created classes appear is definitely a feature but specifically for when `Select all classes` has been enabled. At the moment the app seems to open with `Select all classes` and stay in that state until at least a class is remove from view or `Select all classes` is pressed.

ralfbecher:
Thanks for re-testing on the fixed build. Closing this as fixed by #191, which is on `main` and in 1.17.4.

[...]

PR #322, opened 2026-08-24
feat(viz): only auto-show new entities when the filter shows everything (#194)

### LINKED ISSUE #194: [idea] only auto-add newly created classes when the class filter shows everything
Follow-up to the discussion at the end of #190, requested by @SuperCowProducts.

## Today

`reconcile_class_filter()` (`orionbelt_ontology_builder/app.py`) adds every newly created class to the Visualization's `Filter Classes` selection, regardless of what the user had selected:

```python
newly_added = all_class_set - known
selected_set = (set(selected) | newly_added) & all_class_set
```

This came from #180, where the goal was that creating a class or a restriction must not wipe a narrowed filter. Showing new classes by default was the natural companion to that.

## The request

Auto-adding is right when the user is showing everything, but wrong when they have deliberately narrowed the filter to a set of key classes, for example to restore a specific view of the graph (see #179). In that case a newly created class should not push itself into the curated view.

> Making the newly-created classes appear is definitely a feature but specifically for when `Select all classes` has been enabled.

## Proposed behaviour

Auto-add only when the previous render had every class selected, and tell the user when a new class was created but is hidden by their filter.

1. Restrict the auto-add. No new state is needed: `known` is the previous render's full class set, so `set(selected) == known` already means "the user was showing everything". When the selection was narrower, new classes stay out of it.

2. When a class is created while the filter is narrowed, surface a toast such as `Bicycle created, hidden by your class filter`, ideally with a one-click way to show it. Without this, a user who narrowed the filter earlier and forgot creates a class, sees nothing appear, and lands back on the confusion that opened #190 ("new entities cannot be found in Visualization").

## Notes

- Two tests in `tests/test_viz_class_filter.py` pin the current behaviour on purpose and will need updating: `test_adding_a_class_keeps_the_narrowed_filter` and `test_cleared_filter_still_admits_a_newly_created_class`.
- The `replaced=True` path (load, import, new, undo) must keep showing everything. It is not a narrowing the user chose.
- An emptied filter is a narrowed filter for this purpose, so a cleared selection should stay cleared.

ralfbecher:
Closes #194.

## What changed

Creating a class used to add it to the Visualization node filter whatever the filter was set to. That is right for the default view and wrong for a curated one: a filter narrowed to a handful of key classes is a view the user built, and a class created afterwards should not force its way into it.

`reconcile_filter_selection()` now auto-adds only when the previous render was showing everything. No new state was needed: `known` is the previous render's full entity set, so `set(selected) == known` already answers the question. A narrowed filter, an emptied one included, keeps what it had.

One safeguard came with it: an entity set with nothing in common with the previous one is treated as a replacement. Diffing used to recover that case on its own, because every entity read as new and so got added. Without auto-add, a narrowing from the previous ontology would have survived into an unrelated one and opened it with an empty graph. `test_disjoint_replacement_shows_everything` pins it.

## Saying what was held back

Restricting auto-add makes the creation invisible, which is the confusion #190 opened with, so the page reports it:

- **A toast** naming the entity and the filter hiding it (`Bicycle created, hidden by your class filter`). It is queued in session state rather than raised where it is detected: the hidden-note recompute lower down reruns the page whenever the hidden count changes, which is exactly that render, and a toast raised before a rerun is discarded with the rest of that pass. The queue is drained past that rerun.
- **A `Show new (N)` button** beside `Select all`, present only while there is something to show. It adds those entities to the selection rather than replacing it, so the curated view survives. Its list is pruned on every render, so an entity that is deleted or shown some other way drops out of the offer by itself.

Both apply to the individuals filter too, which goes through the same reconcile.

## Tests

- `tests/test_viz_class_filter.py`: the two tests the issue named are updated, plus the new rule, the disjoint-replacement safeguard, `newly_hidden_uris()` and the toast wording.
- `tests/test_viz_class_paste.py`: `test_a_new_namespace_twin_does_not_unhide_its_sibling` keeps its point (state is keyed by URI, not by label) with the new expectations.
- `tests/test_viz_new_class_hidden.py` (new): drives the page through AppTest, covering the class staying out, the toast, the button and its click, the offer surviving a plain rerun, and a deleted entity dropping out of it.

Driving the Visualization page across renders needed a small shim: AppTest reads every button group's value as a list, but a single-select `st.segmented_control` stores a plain string, so replaying widget states for a second run raises inside Streamlit's own element tree. `_rerun()` hands each group the list form first. That is what the note in `test_viz_delete.py` ("AppTest cannot run the Visualization page twice") was working around.

pytest, ruff and mypy are green. Verified live in the running app: narrowed the class filter to 3 of 4, created `Bicycle` on the Classes page, and the Visualization page came back with the toast, the graph unchanged, and `Show new (1)` in the panel; clicking it brought `Bicycle` in and left the class the user had hidden hidden.


ralfbecher:
Fixed in 199d09a. Confirmed the report: with `known {A, B}` and `selected {A}`, renaming A to A2 left the reconcile with `{A} & {A2, B}` = empty, and `newly_hidden_uris` then reported A2 as freshly created. Renaming the one class you had narrowed the view to emptied the canvas.

The rename notes that keep focus seeds pointing at their entity (issue #275) already carry what is needed, so the filter now follows them before reconciling, in a new `follow_filter_renames()`. Notes are keyed by graph node id, so the current URIs go through `viz_node_id()` to be matched, and each entry in `_FILTER_KINDS` carries the node kind its ids are minted under (individual ids take a prefix, so computing them as classes would have matched nothing).

Selection and known set are re-pointed together. Moving only the selection would leave the new URI absent from `known`, so it would read as newly created and be announced on the very next render.

The notes are peeked, not popped: the focus-seed block further down owns clearing them, and taking them early would put issue #275 back. They are ignored on a replacement, for the same reason that block ignores them.

Tests, at both levels:

- `tests/test_viz_class_filter.py`: the renamed selected class, a renamed hidden class staying hidden, a rename while showing everything, a two-step rename chain, individuals via their own node ids, renamed-then-deleted, and the no-notes and first-render no-ops.
- `tests/test_viz_new_class_hidden.py`: renames through the real path (`rename_class` plus `viz_note_rename`) and checks the class stays selected with no toast and no `Show new` button, plus that the notes are still consumed once.

1563 tests, ruff and mypy green. Verified live too: narrowed the class filter to 4 of 5, renamed `Person` to `Human` on the Classes page, and Visualization came back still holding `1 hidden by the node filter` with `Human` in the selection, no toast and no button.

ralfbecher:
Fixed in 194b937. Correct on both counts: `_viz_new_hidden_{kind}` and the queued `_viz_new_hidden_announce` are per-ontology state, and `_clear_viz_file_session_state()` is where per-ontology state gets dropped on a relink. They now go with the selections and focus seeds already cleared there.

The union at `visualization.py:614` is deliberate (it is what lets the offer survive an unrelated rerun), which is exactly why the list has to be cleared at the switch rather than left for the prune.

Worth noting the neighbouring case does not need anything: a plain ontology swap (load, import, new, undo) reconciles to "everything shown", so the `uri not in _sel_set` prune that runs every render empties the list by itself. Only the relink path, where the new file's saved state is restored as a narrowing, could carry a stale entry through.

`tests/test_viz_file_state.py` gains `test_switching_files_also_forgets_what_was_created_but_hidden`, alongside the existing focus-seed one.

1564 tests, ruff and mypy green.

PR #323, opened 2026-08-24
feat(viz): paste / copy for the focus nodes too (#283)

### LINKED ISSUE #283: [idea] `Paste / copy` for `Focus on one node` mode too

ralfbecher:
Closes #283.

## What changed

Focus mode gets the node filter's restore-a-view box (issue #179). Picking seeds out of a long list is slow and a focus worth setting is worth keeping, so focus mode now takes a pasted list and prints the current one back in the same syntax. The button sits where the filter's does, so it does not move when the panel switches between the two modes.

## The syntax

Seeds are written as the names the picker lists, and read back into the labels the multiselect holds them under.

- A plain name resolves across every focusable kind, and a name two kinds share focuses on both. That is the filter's rule for namespace twins, one level up.
- `Kind:Name` picks one of them, and is what the copy line prints for a shared name. The kind closes its space up (`DataProperty:`), and so does a name the picker has had to tag for its namespace (`zero(fn)`) - neither survives being split out of a pasted line, and the parser closes the pasted text up the same way, so both forms read.
- Full URIs work, as they do in the filter. A SKOS concept's node is keyed by name rather than URI, so it falls back to its label as the identity.

Two small pure helpers carry it, next to the ones the filter's box uses: `build_focus_seed_entries()` turns the picker's targets into entries `parse_filter_text()` and `filter_entry_token()` already understand, and `parse_focus_seed_text()` answers in picker labels.

## A bug in the existing box, fixed with it

While testing the new box I found the filter's own one has been losing its warning. Applying a paste narrows the filter, which changes the "N hidden by the node filter" note, which reruns the page - and the warning was drawn on the pass that reran, so it went with it. A partly matching paste applied silently and never said what it had dropped.

Both boxes now record what they ignored in session rather than warning from the click handler. It survives to the pass the user sees and stands until the next Apply replaces it. The filter's record is per kind, so the individuals box no longer inherits what the classes box ignored. This is the same failure mode as the queued toast in #322, one control over.

## Tests

- `tests/test_viz_focus_paste.py` (new): the syntax (plain names, the shared-name rule, `Kind:Name`, URIs, the tagged form with and without its gap, separators and wrapping), what the copy line writes, an explicit copy-then-paste round trip, and the box on the page - applying a list, a paste matching nothing leaving the focus alone, and a partly matching one applying and still reporting the rest.
- `tests/test_viz_class_paste.py`: page-level regressions for the warning fix, including that it stands until the next Apply.

1587 tests, ruff and mypy green. Verified live: with the focus on `Class: Bicycle`, pasting `Department, Dog, Nonexistent` refocused on `Class: Department` and `Concept: Dog` (a class and a SKOS concept, from one list), left `Ignored, not focusable ...: Nonexistent` under the button, and the copy line then read `Department Dog`.


ralfbecher:
Fixed in 0a4cd57. Confirmed the case is reachable, not theoretical: typing a resource both `owl:Class` and `skos:Concept` puts it in `get_classes()` and `get_concepts()` with one IRI, so the picker lists `Class: Agent` and `Concept: Agent` and the box keyed both on the same URI.

The picker's label is now the entry identity. It is what tells the two apart and what the multiselect holds them under, so `parse_filter_text()` answers in labels directly, and `parse_focus_seed_text()` plus its label map are gone rather than being fixed: with the identity right, the wrapper had nothing left to do.

The IRI rides along as an alias, a new optional `aliases` list on an entry: extra spellings the parser registers without them being the identity. It is the right answer on its own terms, not just a way to keep URI pasting working, since the IRI does name both targets and so focuses on both. Entries that declare no aliases (every node-filter one) take the same path they always did.

New tests: a punned pair keeping both seeds, the kind qualifier still picking one side, the IRI focusing on both, a guard that an alias never becomes the identity, node-filter entries unaffected, and an end-to-end one that builds a genuinely punned ontology, copies the focus off the page and pastes it back.

1593 tests, ruff and mypy green.

Issue #288, opened 2026-08-16
[idea] hotkeys for actions


SuperCowProducts:
For example, when a class is selected:
- `1` -> `Add subclass`
- `2` -> `Add relation`
- `3` -> `Add restriction`
- `4` -> `Add individual`
- `5` -> `Add annotation`
Otherwise:
- `0` -> `Add class`

Issue #276, opened 2026-08-11
[idea] `Alt`+`Click` to Focus on item and remove any other items previously focused


SuperCowProducts:
I often find myself needing `Focus on item` only for one item at a time, so thought something like this would be neat.

Issue #235, opened 2026-08-06
[idea] save `Focus node(s)` after en/disabling `Focus on one node`


SuperCowProducts:
The convoluted system I was brainstorming in https://github.com/ralforion/orionbelt-ontology-builder/issues/224 collapsed into this more targeted concept: when the user enables `Focus on one node`, all nodes in `Filter nodes` are carried over: while it is possible to delete them all and select your preferred one, the user's selection is overwritten on the next switchover. I think it would be more functional to instead restore the user's selection, even in a configuration file so that it persists in the next session, a bit like what already happens with `Filter Nodes` (only there is a list for one and the other state of  `Focus on one node` so they are kept separate, not a general, unified one).

Issue #56, opened 2026-06-27
[idea] `ctrl`-click on node to focus on it in Visualization


SuperCowProducts:
Thought this might be handy when exploring, rather than having to type the name out after seeing it.

PR #334, opened 2026-08-28
feat(viz): a modifier-click on a focused node undoes itself

### LINKED ISSUE #328: [follow-up] When `Focus on one node` is on, if an already-focused node is ctrl-clicked, toggle its visibility
See https://github.com/ralforion/orionbelt-ontology-builder/issues/328#issuecomment-5452310160

<details>

<summary>
original
</summary>

I suppose we might as well make it so that with alt-click every other item in `Focus on one node` is cleared, to mirror the currently-existing behaviour.

</details>

ralfbecher:
Closes #328

Ctrl/Cmd-click adds a node to the focus (#56) and Alt-click focuses on it alone (#276), but clicking a node that was *already* focused did nothing at all, so there was no way out of focus mode except reaching for the checkbox.

## The rule

Each modifier undoes its own action. Only the bold cells change, and **every one of them was a no-op before**, so nothing that did something now does something else.

| | not a seed | one of several seeds | the only seed |
| --- | --- | --- | --- |
| **Ctrl/Cmd-click** (*add*) | add, mode on | **remove it** | **remove it, mode off** |
| **Alt-click** (*set to exactly this*) | focus on it alone | focus on it alone | **mode off, seeds kept** |

Alt-click is deliberately not a general toggle. It is a statement of the target state rather than an accumulate, so turning it into one outside the already-exactly-this-node case would make the outcome depend on state the user cannot see. In the one case it does fire, the graph is visibly showing that node's neighbourhood, so which way the click will go is on screen.

The Alt-click exit keeps the seeds, so re-ticking the checkbox returns you where you were (#235: a focus worth setting is worth keeping). The Ctrl-click removal empties them, because you said remove it.

## The trap this had to avoid

`render_visualization` backfills an arbitrary first label when focus mode is on with no seeds:

```python
if not saved_seeds:
    saved_seeds = [focus_labels[0]]
```

So the mode has to go off in the same breath as the seeds empty. Passing through that state on any later render would jump the graph to a class nobody picked. `focus_seeds_after_request` now returns `(seeds, focus_on)` and the caller sets both together.

## Live check

Driven through the component's own click handler in the running app:

| step | focus mode | seeds |
| --- | --- | --- |
| ctrl-click Person | on | Human |
| ctrl-click Organization | on | Human, Organization |
| ctrl-click Human again | on | Organization |
| ctrl-click Organization (last) | **off** | emptied, nothing backfilled |
| alt-click Role | on | Role |
| alt-click Role again | **off** | kept |
| re-tick the checkbox | on | Role |

## Note on the issue

@SuperCowProducts' follow-up comment suggested making ctrl *and* alt per-node toggles, with a keyboard shortcut for the mode. This takes the ctrl half, keeps alt as the "exactly this" gesture for the reason above, and leaves the shortcut to #288 (hotkeys), which is the better home for it.

Tests rewritten in `tests/test_viz_focus_click.py` (9). Full suite (1626), ruff and mypy green.

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -218,6 +218,7 @@ from .ui import (  # noqa: F401
     set_active_language_pack,
     set_flash_message,
     show_message,
+    viz_apply_focus_click,
     viz_auto_show_new_toggled,
     viz_drop_focus_seeds,
     viz_filter_changed,
--- a/orionbelt_ontology_builder/ui.py
+++ b/orionbelt_ontology_builder/ui.py
@@ -1275,11 +1275,11 @@ def focus_seeds_from_selection(selected_classes, class_count):
     return labels
 
 
-def focus_seeds_after_request(seeds, label, replace=False):
+def focus_seeds_after_request(seeds, label, replace=False, focus_on=False):
     """What a modifier-click on a graph node leaves the focus as.
 
-    Returns ``(seeds, focus_on)`` — the new seed list and whether focus mode
-    should still be on.
+    ``focus_on`` is whether focus mode is on *now*; returns ``(seeds, focus_on)``
+    — the new seed list and whether it should be on after the click.
 
     Ctrl/Cmd-click adds the node to whatever is focused already (issue #56),
     which is what building a neighbourhood up out of several nodes needs.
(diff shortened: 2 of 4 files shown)


SuperCowProducts:
In fact @ralfbecher, I was thinking we could avoid `The trap this had to avoid` entirely on https://github.com/ralforion/orionbelt-ontology-builder/issues/335. May streamline development on this if you consider the suggestion valid.

PR #343, opened 2026-08-29
feat(sparql): read-only SPARQL query console

ralfbecher:
Adds the SPARQL page: SELECT, ASK, CONSTRUCT and DESCRIBE over the loaded ontology, with the ontology's own prefixes pre-declared (so `owl:Class` and `:Person` work without a PREFIX line), eight worked examples, and CSV/Turtle download of results.

## Read-only

`prepareQuery` accepts only the four query forms, so every update form fails it. A failure is re-parsed as an update to tell two different problems apart: "this console is read-only" and "you have a typo" get different messages. The store the query runs against also raises on any write, so the guarantee holds by construction rather than by argument. Edits keep going through the editor pages, where they are checkpointed and can be undone.

## Bounded, in two layers

A row cap and a deadline on result iteration catch the common runaway, a query that streams more rows than anyone wants to see.

The layer that actually saves the app is a deadline checked inside a read-through store proxy. rdflib's `evalBGP` pulls every candidate solution through `ctx.graph.triples()` and recurses per binding, so a deadline at the store is checked continuously throughout evaluation, including while an `ORDER BY` / `GROUP BY` / `DISTINCT` above it consumes the whole join before yielding anything.

Measured on a 300-triple graph:

| Case | Result |
| --- | --- |
| 3-way cartesian + `ORDER BY`, deadline on result iteration only | no first row at all after 8s |
| Same query, deadline inside `store.triples()` | stopped at 3.05s |
| Realistic query, 6k-triple graph, proxy installed | 72.1ms vs 72.3ms plain |

Staying in-process is the point. A subprocess would mean forking from a Streamlit ScriptRunner thread, which is a multi-threaded process; CPython has deprecated that since 3.12 ("use of fork() may lead to deadlocks in the child") and the Docker image ships 3.14. The proxy needs no fork, no signals and no main thread, so the server and the desktop build behave identically.

## The gap this does not close

A query with no triple patterns at all, such as a cross product of large `VALUES` blocks, never touches the store, so nothing trips. That shape is rejected before execution rather than left as a hole. `SERVICE` is rejected too: it would have the server open a connection to a host named in the query.

## Notes

- The wrapper graph inherits the source graph's `identifier`. Without that it addresses an empty context and every query silently returns nothing, which looks exactly like a clean pass.
- URI shortening is done against the graph's own prefix list rather than through rdflib's `normalizeUri`, which binds generated `ns1:` prefixes as a side effect and would leave them on the user's ontology. There is a test for that.
- The row limit is a `number_input` rather than a dropdown, matching the time limit beside it and staying clear of the plain-selectbox guard in `test_clearable_required_dropdowns.py`.

## Testing

46 new tests across `tests/test_sparql.py` (engine) and `tests/test_sparql_ui.py` (page), plus the new page in the smoke matrix. Full suite 1709 passed, ruff and mypy clean.

Verified live in the running app: examples, prefixed results, the read-only refusal (graph untouched at 45 triples), and a five-way cartesian under `ORDER BY` stopped at 3.06s against a 3s budget. That live pass found one bug, now fixed: a timed-out query with no rows also reported that it "matched nothing", which was the opposite of what happened.

The version is left at 1.25.0. Bumping to 1.26.0 is the release step and is yours to call.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,6 +25,7 @@ orionbelt_ontology_builder/
 ├── cli.py, desktop.py  # `orionbelt-ontology-builder[-desktop]` console entry points
 ├── ontology_manager.py # Core OWL/SKOS engine on rdflib (OntologyManager, UndoManager)
 ├── templates.py        # Built-in templates, upper & reference ontologies
+├── sparql.py           # Read-only SPARQL execution (deadline + row cap)
 ├── samples/            # Bundled ontologies (gist, gUFO, FOAF, PROV-O, …)
 ├── lib/                # The graph component (vendored vis-network)
 ├── assets/             # Logos, screenshots
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -60,7 +60,7 @@ pytest                                    # testpaths/pythonpath configured in p
 pytest tests/test_classes.py -q           # a single file
 ```
 
-Dependencies: streamlit, rdflib, owlrl, networkx. Python >= 3.12.
+Dependencies: streamlit, rdflib, owlrl, networkx, streamlit-ace. Python >= 3.12.
 
 The floor is 3.12 because that is the oldest runtime CI keeps green, and
 because both deployment targets are at or above it (Streamlit Community
--- a/README.md
+++ b/README.md
@@ -250,7 +250,8 @@ Imports on an empty ontology go straight through. Otherwise you get a review pan
 
 A read-only query console over the loaded ontology: SELECT, ASK, CONSTRUCT and
 DESCRIBE, with the ontology's own prefixes already declared so `owl:Class` and
-`:Person` work without a PREFIX line. Results download as CSV (SELECT) or
+`:Person` work without a PREFIX line. The editor highlights SPARQL syntax and
+follows the app's light/dark theme. Results download as CSV (SELECT) or
 Turtle (CONSTRUCT/DESCRIBE), and eight worked examples are one click away.
 
 Updates (INSERT, DELETE, LOAD, CLEAR, DROP) are refused: edits go through the
@@ -519,7 +520,7 @@ orionbelt-ontology-builder/
 └── tests/                              # pytest suite
 ```
 
-Dependencies: streamlit, rdflib, owlrl, networkx.
+Dependencies: streamlit, rdflib, owlrl, networkx, streamlit-ace.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -246,6 +246,18 @@ Imports on an empty ontology go straight through. Otherwise you get a review pan
 - SKOS checks (see above)
 - RDFS and OWL-RL reasoning via owlrl
 
+### SPARQL queries
+
+A read-only query console over the loaded ontology: SELECT, ASK, CONSTRUCT and
+DESCRIBE, with the ontology's own prefixes already declared so `owl:Class` and
+`:Person` work without a PREFIX line. Results download as CSV (SELECT) or
+Turtle (CONSTRUCT/DESCRIBE), and eight worked examples are one click away.
+
+Updates (INSERT, DELETE, LOAD, CLEAR, DROP) are refused: edits go through the
+editor pages, where they are checkpointed and can be undone. Every query runs
+under a row limit and a wall-clock time limit, so a runaway query is stopped
+rather than taking the app with it.
+
 ### Visualization
 
 Interactive vis-network graph with class filtering, configurable node limits, click-to-navigate into the editor, Ctrl/Cmd-click a node to add it to the "Focus on one node" selection (narrowing the graph to its neighbourhood) or Alt-click to focus on it alone, hierarchy tree view, and statistics charts.
@@ -483,6 +495,7 @@ the 200 MB default; raise the value only when self-hosting with enough memory.
 | **SKOS Vocabulary** | Concept schemes, concepts, hierarchy, SKOS validation          |
 | **Import / Export** | File import with merge review, export, new ontology, templates |
 | **Source**          | Live Turtle source view of the ontology                        |
+| **SPARQL**          | Read-only SPARQL console with row and time limits              |
 | **Validation**      | Ontology validation and OWL reasoning                          |
 | **Visualization**   | Interactive graph (OWL + SKOS), hierarchy tree, statistics     |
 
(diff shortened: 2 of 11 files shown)


ralfbecher:
Stress-testing the time limit in the running app turned up something that strengthens the case for the store deadline, now in a08a6f7.

rdflib's `evalUnion` is not a generator. It appends both branches into a list and returns it:

```python
def evalUnion(ctx, union) -> List[Any]:
    branch1_branch2 = []
    for x in evalPart(ctx, union.p1):
        branch1_branch2.append(x)
    for x in evalPart(ctx, union.p2):
        branch1_branch2.append(x)
    return branch1_branch2
```

So every query containing a `UNION` is fully evaluated before its first row exists. On the 46-triple test ontology, `graph.query()` alone took 1381ms on a three-branch union before returning anything.

That makes `UNION` a much more common trigger for "blocks before the first row" than `ORDER BY` or `GROUP BY`, which is what the original write-up cited. It is also the shape any shortest-path query takes in plain SPARQL, since hop counts have to be enumerated as union branches. The row cap cannot bound it at all, because there are no rows to cap until evaluation finishes.

Verified in the app on a five-branch union: stopped at 10.14s against a 10s budget, and 3.03s against a 3s one, zero rows both times, the app responsive afterwards. Test added for the union case.

ralfbecher:
Good catch, fixed in 2e2ed9a. Confirmed the reproduction, and the blast radius is wider than the report.

The finding says `FROM NAMED` produces a generic dataset error. It does not. On a one-triple graph, every dataset shape silently returned the full graph:

| Query | Result before the fix |
| --- | --- |
| `SELECT * WHERE { ?s ?p ?o }` | 1 row |
| `SELECT * FROM <http://no-such.example/g> WHERE { ?s ?p ?o }` | same 1 row |
| `SELECT * FROM NAMED <http://no-such.example/g> WHERE { ?s ?p ?o }` | same 1 row |
| `CONSTRUCT { ?s ?p ?o } FROM <http://no-such.example/g> WHERE { ?s ?p ?o }` | same 1 triple |

So `FROM NAMED` was not erroring, it was misleading in exactly the same way as default `FROM`, and `ASK` and `DESCRIBE` behave the same. All five shapes are now refused and covered by a parametrised test.

The fix is one name in the guard that already walks the algebra for `SERVICE`: rdflib exposes the clause as a `DatasetClause` node, so `_algebra_guard` rejects it with a `QueryNotAllowed` that says the console always queries the loaded ontology and the clause would be ignored rather than honoured.

Worth stating why refusing beats accepting-and-warning here. The query is accepted, rows come back, and they look right, so `FROM <somewhere-else>` hands back this ontology's data under another graph's name with nothing to signal the substitution. A caption nobody reads is not a fix for that.

Full suite 1723 passed, ruff and mypy clean.

One unrelated thing I noticed while reproducing, not worth folding into this commit: `SELECT *` column order is not stable across processes. rdflib derives the projection for `*` from a set, so hash randomisation reorders it between runs, while an explicit `SELECT ?s ?p ?o` is stable. Header and values are always zipped from the same vars so nothing is mislabelled, but repeated CSV exports of the same `SELECT *` can differ in column order. Happy to pin it if that is worth a follow-up.

ralfbecher:
Followed up on the column-ordering note from the previous comment: pinned in 2534c01.

`SELECT *` now takes the order the query first names each variable, read off the algebra's BGP, VALUES and BIND nodes rather than the query text, so comments and string literals cannot affect it. `SELECT * WHERE { ?s ?p ?o }` gives `s, p, o` on every run.

The part worth recording is what the first attempt got wrong. The plan was to order every projection by first appearance, on the reasoning that an explicit projection lists its variables in the SELECT clause and would therefore be a no-op. That is true of the query *text* and false of the algebra, which keeps no record of the SELECT clause at all. Ordering everything against the pattern rewrote the user's projection:

| Query | First attempt | Correct |
| --- | --- | --- |
| `SELECT ?o ?s WHERE { ?s ?p ?o }` | `s, o` | `o, s` |
| `SELECT ?p (COUNT(?s) AS ?n) …` | `n, p` | `p, n` |

So whether the user wrote `*` is now read off the parse tree. `prepare()` calls `parseQuery` and `translateQuery` directly instead of `prepareQuery`, which is exactly those two calls and throws away the tree that answers the question — one parse, no extra cost.

Two tests guard it. Column order is asserted equal across four `PYTHONHASHSEED` values in subprocesses, because one interpreter has a single hash seed for its lifetime and an in-process test cannot see this bug at all. And each value is asserted to land under its own column, since a reordered header over unreordered cells would be worse than the instability it replaces.

Scope note: this is confined to `run_query`'s column and cell construction. The page passes `columns`/`rows` straight through, so it is covered by the engine tests plus the existing page tests, not by a fresh browser check.

Full suite 1730 passed, ruff and mypy clean.

ralfbecher:
Confirmed and fixed in 8700b9e. Both repros were exact, and the diagnosis was right down to the mechanism: the algebra puts the operator that introduces a variable *above* the pattern it reads from, so a pre-order walk reaches BIND's variable before the pattern preceding it in the source.

Took the second of the two suggested fixes rather than the first, because special-casing Extend inverts a different case. `BIND(1 AS ?first) ?s ?p ?o` is legal, and visiting Extend's `p` child before recording `var` would push `?first` behind the pattern it precedes:

| Query | Before | Special-case Extend | Parse tree (shipped) |
| --- | --- | --- | --- |
| `{ ?s ?p ?o . BIND(?o AS ?x) }` | `x, s, p, o` | `s, p, o, x` | `s, p, o, x` |
| `{ VALUES ?z { 1 } BIND(2 AS ?a) }` | `a, z` | `z, a` | `z, a` |
| `{ BIND(1 AS ?first) ?s ?p ?o }` | `first, s, p, o` | `s, p, o, first` | `first, s, p, o` |

The same inversion would also be waiting in any other operator that binds above its child, so the order now comes from the parse tree, which mirrors the source and has no inversion to correct. Not from the raw query text either: a variable named in a comment or inside a string literal is not a binding, and the parser has already told them apart — there is a test for exactly that.

This cost nothing extra. `prepare()` already parses in order to answer whether the user wrote `SELECT *`, so the tree was in hand.

New coverage for the gap you identified at tests/test_sparql.py: BIND after a pattern, BIND before one, VALUES with BIND, OPTIONAL, UNION, and the comment/string case. The existing cross-hash-seed test now runs against a BIND query too.

Full suite 1736 passed, ruff and mypy clean.

Issue #189, opened 2026-07-25
[glitch] `Open full editor` button briefly appears when (de)/selecting entity in fullscreen


SuperCowProducts:
I think we'll want to implement a switch for it too - not of the fact it's selected, but the popup appearing as a consequence (or maybe we can force fullscreen on top of *everything* - if we did already strange it appears for a second then vanishes though).

Issue #172, opened 2026-07-22
[suggestion] Discussions tab for repo


SuperCowProducts:
Something like Discussions or anything forum-like would be great for sharing ideas on using OrionBelt. In the future it can also help foster community and get beginners up-to-speed quicker.

For example, I recently got AI to build me a tool that found the shortest path between 2 classes of my ontology, which worked :):

<details>

<summary>
Python script
</summary>

```python
#!/usr/bin/env -S uv run --script
# /// script
# requires-python = ">=3.12"
# dependencies = ["rdflib","networkx"]
# ///
from rdflib import Graph, URIRef
import networkx as nx

# --- Load the .ttl file ---
g = Graph()
g.parse("ontology.ttl", format="turtle")  # Replace with your file path

# --- Create a NetworkX graph ---
nx_graph = nx.Graph()

# --- Add edges based on specific properties ---
# Example: Only use rdfs:subClassOf or custom properties like :relatedTo
for s, p, o in g:
    # Option 1: Use only rdfs:subClassOf (common for class hierarchies)
    if p == URIRef("http://www.w3.org/2000/01/rdf-schema#subClassOf"):
        nx_graph.add_edge(s, o)

    # Option 2: Use a custom property (e.g., :relatedTo)
    # elif p == URIRef("http://example.org/ontology#relatedTo"):
    #     nx_graph.add_edge(s, o)

# --- Define your entities (replace with actual URIs from your graph) ---
entity1 = URIRef("http://example.org/ontology#sinusoid")
entity2 = URIRef("http://example.org/ontology#sinusoidal-V")  # Replace with your target

# --- Debug: Print all nodes and edges ---
print("Nodes in the graph:", nx_graph.nodes())
print("Edges in the graph:", nx_graph.edges())

# --- Check if entities exist ---
if entity1 not in nx_graph or entity2 not in nx_graph:
    print("Error: One or both entities not found in the graph.")
    print("Available nodes:", nx_graph.nodes())
else:
    # --- Find the shortest path ---
    try:
        shortest_path = nx.shortest_path(nx_graph, source=entity1, target=entity2)
        print(f"\nShortest path between {entity1} and {entity2}:")
        print(" -> ".join(shortest_path))
    except nx.NetworkXNoPath:
        print("No path found.")
```

</details>

[...]

PR #352, opened 2026-08-30
feat(viz): shortest path between two entities (#176)

### LINKED ISSUE #176: [idea] Shortest path between two entities in the graph
Add a tool that finds and highlights the shortest path between two entities (classes, individuals, or SKOS concepts) in the ontology graph.

## Motivation

Suggested by @SuperCowProducts in #172, with a working proof of concept: a small `rdflib` + `networkx` script that loads the ontology and returns the shortest path between two class URIs along `rdfs:subClassOf` (and optionally other properties). Both libraries are already app dependencies, so this fits naturally.

## Sketch

- Pick a source and a target entity (two searchable selectboxes, similar to the new "Find entity" picker).
- Build a `networkx` graph from the ontology's edges. Options worth considering:
  - which edge types to include (subClassOf only, all object properties, everything);
  - directed vs undirected (undirected is friendlier for "how are these two related").
- Compute the shortest path with `networkx.shortest_path`; report "no path" cleanly when disconnected.
- Surface the result:
  - as text (the chain of hops), and
  - ideally by highlighting the path in the Visualization graph (colour the path nodes/edges, or focus on the sub-path), reusing the existing graph component.

## Where

Either a small section on the Visualization page (it already has the graph and the find/focus machinery) or the Advanced page. Visualization is probably the better home since the payoff is visual.

## Notes

- Keep it robust for large graphs (bound the work, handle disconnected components).
- The reference script from #172 is a good starting point for the graph-building and pathfinding logic.

ralfbecher:
Closes #176.

Pick a source and a target on the Visualization page and the graph reports the shortest chain of links between them, written out hop by hop and highlighted on the canvas.

Suggested by @SuperCowProducts in #172, with a working proof of concept in discussion #178.

## What it does

A new "Shortest path between two entities" panel under the Find / Focus row:

- two pickers, listing the same entities as Find (classes, individuals, SKOS concepts);
- the result as one line, naming each link and the direction it was asserted in, e.g. `Class: Role ←subClassOf— Class: Department ←hasDepartment— Class: Organization`;
- the links along the path repainted on the canvas, and their nodes given a path border;
- "Show only this path", which turns the path into the focus.

## How it is built

**The search runs in the engine, over the whole rdflib graph**, not over the assembled graph. The canvas builds at most `GRAPH_MAX_NODES` (500), so a search over what is drawn would report "no path" on exactly the ontologies where the question is worth asking. `OntologyManager.build_path_graph()` and `find_shortest_path()` are pure enough to unit test directly, and `bfs_path()` is a plain function over an adjacency.

**Which kinds of link it may walk comes from the display toggles** (`path_edge_kinds`), so the path it finds is always made of edges the canvas draws. Each condition mirrors the guard the graph builder puts on the edges it adds; a path over a relation the canvas is not drawing could not be highlighted and would read as the highlight being broken.

**Undirected**, since the question is "how are these two related". The direction each link was asserted in is kept, so the text can still read it back the way the ontology states it.

**A data property is not walkable**: its node hangs off its domain class and leads nowhere else, so it can neither be a step in a path nor a useful endpoint. It is left out of both pickers.

## Two things worth a reviewer's eye

The path can still name entities the current view has no node for (past the cap, or hidden by a filter). When that happens the graph says so where it already explains why it is smaller than the ontology, and "Show only this path" is the way back: focus mode assembles past the render cap and prunes to its seeds.

A repainted edge keeps the colour of its own kind in `basecolor`, and the canvas legend counts both. Without that, highlighting the only `subClassOf` edge on screen took `subClassOf` off the legend.

## Notes

- `_graph_ver` bumped to 21, so a session holding a pre-#176 payload does not keep serving a graph with no highlight in it. `test_the_graph_cache_version_moved_with_the_pruning` pinned the number exactly, which every future bump would break; it now asserts the floor that was actually its point.
- Not in scope, left on #176: k-shortest paths (discussion #178) and connecting n entities (SuperCowProducts' follow-up comment), which is a Steiner-tree problem rather than a shortest path.

## Testing

- `tests/test_shortest_path.py` (24) covers the engine and the pure page helpers.
- `tests/test_viz_shortest_path.py` (18) drives the real `render_visualization` through AppTest and inspects the graph payload: the repaint, what the panel says in each case, and that the panel keeps a constant element count (an element that comes and goes above the graph component re-creates its iframe, issue #189).
- 1796 pass; ruff and mypy clean.
- Verified live in the running app with Playwright: the text, the coloured edges, the legend, and "Show only this path".


SuperCowProducts:
Can't wait to try this out!

Issue #356, opened 2026-08-31
[bug] Your app is having trouble loading the streamlit_ace.streamlit_ace component


SuperCowProducts:
The new `SPARQL` section is great to have - after all it's what structuring data is all about in the first place, as it scales. Not sure what's going on when this happens after trying out the predefined queries - in any case feel free to play with the ontology I'm using, I built it for personal use so I'm not sure how useful it could be to others, which is why I'm not hosting it anywhere, but there's nothing in it stopping it from being shared, in fact I'm glad it at least provides a real-world use case: [ontology.ttl.txt](https://github.com/user-attachments/files/31634054/ontology.ttl.txt).

<img width="2560" height="1600" alt="Image" src="https://github.com/user-attachments/assets/865ea576-4b06-4fb9-9583-4cd8b24782cf" />

In any case, the query definitely produces output, so maybe this is mainly just 'background app noise' and doesn't actually affect any functionality.

ralfbecher:
Thank you for the report, and for the ontology - importing it (1,008 classes, 5,698 triples) is how this got pinned down, and every example query runs against it.

Your read is right: it is noise, and the query output is unaffected. Here is what causes it.

That banner is Streamlit's own message when a component does not report ready within three seconds. Two things line up to provoke it, and both do on every example you pick:

**The editor is thrown away and rebuilt.** The Ace editor takes its text as *initial* content and ignores it afterwards, so the only way to show you a different query is to mount a fresh copy of it. A fresh copy means a new iframe, which starts from an empty document and has to load the whole editor again.

**Your browser is not allowed to keep what it just downloaded.** Streamlit serves component files with `cache-control: public` and nothing else - no expiry, no `ETag`, no `Last-Modified`. With no lifetime and nothing to revalidate against, the browser goes back to the network every time. Measured: 286 KB gzipped (1.06 MB unpacked), per example.

On a desktop connection that finishes in well under three seconds and you never see it. On mobile data it does not, which fits where you are seeing it.

Neither half is ours to fix: the files belong to `streamlit-ace` (last released December 2021), and the headers come from Streamlit's own static file handler.

So #369 removes the download instead. There is now a **Plain editor** toggle next to the Query label: it swaps the highlighted editor for a plain text box - no iframe, nothing to load, nothing to reload when you pick an example. You lose the syntax colouring; you gain a page that behaves on a slow link. The highlighted editor stays the default for anyone not affected.

[...]

PR #377, opened 2026-08-31
fix(viz): say when a path is only half drawn, even on a capped graph

ralfbecher:
Reported against v1.27.1: a highlighted path on a large ontology comes out in disconnected pieces.

## What is actually happening

Nothing is wrong with the highlight. Reproduced on the ontology from #356, asking for a path from `right-hand rule` to `json`:

- the path is **16 hops, 17 entities**, all `subClassOf`;
- the view holds **500 of 1008** classes, because of the node cap;
- **9 of the 17** entities on the path are among the ones it does not hold;
- so only the **4** links with both ends drawn can be ringed, and the path looks broken.

## Why nothing said so

The page has a line for exactly this case, and it could never appear:

```python
if _path_missing and not graph_notice:
```

A graph at the node cap has always already set the generic *"N of M classes are not drawn"*, so the guard was never satisfied. The one message that explains the pieces - and the only one that names the way out, `Focus on this path` - was suppressed by the vaguer message that shares its cause.

It is now said *instead of* that line rather than only in its absence, on the grounds that it is the more specific of the two and it is about what the user is looking at right now.

## Tests

Two, and the first fails without the change: a path that cannot fit under the cap reports itself, and a path drawn in full leaves the graph's own message alone.

`pytest` 1859 passed, `ruff` and `mypy` clean.

## Not in this PR

The better answer, suggested while reviewing this: keep the path's entities *in* the graph by giving them priority within the cap, dropping unrelated nodes to make room, the way the Find target already survives the cap (#234). That is a change to the builder's gates rather than to a message, and it wants its own PR - this one stops the current release lying by omission.

### DIFF
--- a/orionbelt_ontology_builder/views/visualization.py
+++ b/orionbelt_ontology_builder/views/visualization.py
@@ -1622,8 +1622,11 @@ def render_visualization():
         # focus mode stopped charging an annotation a hop (issue #272), so a
         # cached focus payload is one built under the old pruning. 21: nodes and
         # edges along a shortest path are repainted (issue #176), which a cached
-        # pre-#176 payload has none of.
-        _graph_ver = 21
+        # pre-#176 payload has none of. 22: the notice cached with the payload
+        # now reports a path that is only half drawn, where before the generic
+        # cap line suppressed it — a cached v21 payload carries the suppressed
+        # one and would keep showing it until something unrelated evicted it.
+        _graph_ver = 22
         # Include a mutation counter that bumps on every checkpoint / undo / redo,
         # so any change to the ontology — even one that preserves triple count —
         # invalidates the cached graph data and the iframe re-renders.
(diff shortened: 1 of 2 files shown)


ralfbecher:
Both findings were right, and the first is worse than the report says.

**P2 — the override took every notice, not just the generic one.** Three of the four notices it could clobber are focus-specific, and one of them (`visualization.py`, "Nothing to focus on") explains an **empty canvas** - replacing that with "use Focus on this path" would have been nonsense on a graph showing nothing at all. The repro in the review is the other half of it: focused on exactly the path, and told to focus on it.

Fixed by marking which notice is in the slot. The generic class-cap line sets a flag; the focus messages clear it; the path line replaces the slot only when it is empty or holds the cap line. Where a focus has something to say, it keeps saying it.

The review's own scenario is now a test - cap of two, focus seeded with the path, path still not drawable - and it fails against the unconditional override.

**P3 — cache version.** Correct, and it is the same trap the file's own history records at versions 19 and 21. The notice rides in `last_graph_data`, so changing what it says changes what a cached payload means: `_graph_ver` is 22, with the reason written next to the others.

`pytest` 1871 passed, `ruff` and `mypy` clean.

PR #379, opened 2026-08-31
feat(viz): a highlighted path is drawn whatever the node cap (#378)

### LINKED ISSUE #378: [idea] keep a highlighted path in the graph, past the node cap
Follow-up to #371's review, and the reason for #377.

## What happens now

Ask for a path on an ontology larger than the node cap and it comes out in disconnected pieces. On the ontology attached to #356, `right-hand rule` to `json`:

- the path is 16 hops, 17 entities, all `subClassOf`;
- the view holds 500 of 1008 classes;
- 9 of the 17 entities on the path are among the ones it does not hold;
- so 4 of the 16 links can be ringed, and the rest of the path is not on screen at all.

#377 makes the page say so, and points at **Focus on this path**. That is a workaround: the user asked a question, got an answer, and has to run a second command to see it.

## What this asks for

The entities on a highlighted path are drawn, whatever the cap. The precedent is #234: the entity picked in **Find** already survives the cap, because `prioritise_find_target` builds it first and each per-kind gate lets that one target past its own budget check.

## Why it is not just a bigger `or`

There are eight cap checks in the builder today (classes, data properties, individuals, annotations twice, SKOS, triples). The Find target is let past each of them one at a time, and every gate added since has had to remember to do it. Threading a second exception the same way is how that gets worse - what this wants is **one** "always include these" set that the gates consult, with Find and the path both in it.

Then the budget question, in increasing order of ambition:

1. **Go over the cap by the path's size.** 17 nodes on a 500 budget is immaterial to the browser, and it is exactly the trade #234 already made for one node. Simple, and it makes the path always whole.
2. **Evict to stay at the cap.** Drop nodes to pay for the path's. Needs a policy for what goes first, and a reason to believe it is better than (1).
3. **Prefer connected nodes when the cap bites.** At 500 nodes a large ontology draws many isolated ones that say nothing about the path. This would help every capped graph, not only this case - but it changes what every capped graph looks like, so it wants its own issue, its own before-and-after, and is not part of this one.

(1) is the proposal here. (2) and (3) are noted so the next reader knows they were considered.

## Notes

- The path is known before the build (the panel renders above it), and the graph cache key already includes the path, so a rebuild on a path change is expected behaviour rather than a new cost.
- Focus mode assembles past the cap and prunes to its seeds, which is why **Focus on this path** works today; this issue is about not having to.

ralfbecher:
Closes #378.

## The problem

Asking for a path on an ontology past the node cap answered the question and then drew the answer with its middle missing. On the ontology from #356, `right-hand rule` to `json`:

| | before | after |
|---|---|---|
| path entities drawn | 8 of 17 | **17 of 17** |
| links ringed | 4 of 16 | **16 of 16** |
| nodes drawn in total | 500 | **500** |

The cap is unchanged: the path's classes are built first, so they displace unrelated ones rather than inflating the graph.

## How

The builder already knew how to protect one entity from the cap - the entity picked in **Find** is ordered first and let past each per-kind budget check (#234). That exception was threaded through the gates one at a time, and every gate added since has had to remember it.

So this does not add a second exception beside the first. Both become **one pinned set of node ids** that the gates consult once:

- `prioritise_find_target(items, node_id_of, find_id)` -> `prioritise_pinned(items, node_id_of, pinned)`, taking a set (a bare id still works);
- `_find_kind_is(prefix)` -> `_pinned_kind_is(prefix)`;
- the eight `!= _find_id` escapes -> `not in _pinned_ids`.

`_pinned_ids` is `{the Find target} | {every entity on the highlighted path}`. The path is known before the build - the panel renders above it - and the graph cache key already includes it, so a path change rebuilding the graph is existing behaviour.

Where a path is longer than the cap, the graph can exceed the cap by what it pins. That is the trade #234 already made for one node, and the issue's option (1).

The node-filter escape comes along with the cap escape: a class the filter hides is drawn when it is on the path, for the same reason it is drawn when it is the Find target - you asked to see it.

## Tests

Two new. The first fails on unmodified source: a path that cannot fit under the cap is drawn whole and ringed end to end. The second guards the other direction - pinning buys the path its own nodes, not a bigger graph, so an unrelated class stays out.

`pytest` 1859 passed, `ruff` and `mypy` clean.

## Relationship to #377

#377 makes the page *say* when a path is only half drawn. With this merged that message becomes the fallback rather than the common case - it still fires where pinning cannot help, such as a path through an entity type that is switched off entirely.

### DIFF
--- a/orionbelt_ontology_builder/app.py
+++ b/orionbelt_ontology_builder/app.py
@@ -197,7 +197,7 @@ from .ui import (  # noqa: F401
     path_nodes,
     persist_autosave,
     persist_language_packs,
-    prioritise_find_target,
+    prioritise_pinned,
     prune_reused_focus_seeds,
     reconcile_filter_selection,
     render_add_annotation,
--- a/orionbelt_ontology_builder/ui.py
+++ b/orionbelt_ontology_builder/ui.py
@@ -4304,31 +4304,45 @@ VIZ_NODE_PANEL = "Node options"
 FOCUS_BUILD_MAX_NODES = 5000
 
 
-def prioritise_find_target(items, node_id_of, find_id):
-    """Build the entity picked in Find first, so the cap cannot drop it (#234).
+def prioritise_pinned(items, node_id_of, pinned):
+    """Build the entities in ``pinned`` first, so the cap cannot drop them.
 
     Each loop stops at a fixed node budget, in list order, so an entity far
-    enough down was never drawn. The Find dropdown lists every entity regardless
-    of what was drawn, so picking one of those left the viewer nothing to centre
-    on. It then dropped the camera pin, and dropping the pin re-enables the
-    post-layout auto-fit, so the graph visibly reframed while the entity the user
-    asked for was missing: movement, and no result, which is exactly what the
-    issue describes.
+    enough down was never drawn. The pickers list every entity regardless of
+    what was drawn, so an entity picked from one of them could be missing from
+    the graph that is supposed to answer for it.
+
+    Two things are pinned. The entity picked in **Find** (issue #234): picking
+    one the cap had cut left the viewer nothing to centre on, it dropped the
+    camera pin, and dropping the pin re-enabled the post-layout auto-fit — so
+    the graph visibly reframed while the entity the user asked for was missing.
+    And every entity on a **highlighted path** (issue #378): a path whose middle
+    is not drawn is ringed in disconnected pieces, which reads as a broken
+    highlight rather than as a view that does not hold all of it.
 
     Ordering alone is enough for classes, which are built first and so still fit
     inside the budget. It is not enough for the kinds after them: by the time
-    those loops run the budget is spent, so each also lets the target itself past
-    its own cap check. That costs at most one node over the cap, since only the
-    single entity the user named is let through. One extra node is immaterial to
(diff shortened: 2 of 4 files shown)


ralfbecher:
Both right, and the first one exposed that the PR's filter-escape claim was only half true.

**P2 — the pin never reached the class loop.** The inner filter check honoured pinned ids, but the loop's own gate did not: `selected_classes or _find_is_class`. A pinned class satisfies neither, because class node ids are bare uids with no kind prefix - `_pinned_kind_is` cannot see them, and only the Find target had a flag of its own. With the filter emptied the graph came out with no nodes at all, exactly as reported.

The gate now tests the pinned ids against the class list, which is what it takes to recognise a class id. `_find_is_class` is gone rather than left beside it: the Find target is pinned too, so the new condition subsumes it, and a flag that exists for one caller is how the next gate forgets about the other one - which is the whole reason this PR consolidated the Find exception in the first place.

The test is your repro: filter emptied, path picked, and the graph holds exactly `{A, B, C}` with both links ringed. It fails against the old gate.

**P3 — cache version.** Correct: a v22 payload is a graph built without the path's nodes in it. `_graph_ver` is 23, with the reason written beside the others.

`pytest` 1874 passed, `ruff` and `mypy` clean.

PR #380, opened 2026-09-01
fix(viz): the graph no longer flashes white when a panel is toggled (#372)

### LINKED ISSUE #372: [UX] avoid flashing when toggling `Display options` and `Path finder`
I was on dark mode, so I presume it's the white background popping up for a brief second.

ralfbecher:
Closes #372.

## What was flashing

Not what I first assumed. Sampling the component's background colour **every frame** across a toggle of Display options, on unmodified `main`:

```
rgb(14,17,23) → transparent → unreadable → rgb(255,255,255) → rgb(14,17,23)
                                            ^^^ three frames of white
```

Streamlit replaces the component's *document* on those toggles, and the viewer's stylesheet opens with:

```css
html, body { margin: 0; padding: 0; overflow: hidden; background: #fff; }
```

So a fresh document paints white and holds it until the theme arrives over the round trip from Python. In a dark app that is the flash, and it lands on two switches people toggle constantly.

## The fix

The viewer paints the colour it last drew, before anything else is parsed. `applyTheme` keeps it in `sessionStorage` (per session, per origin) and a script in the head turns it into a stylesheet - a stylesheet rather than a style property, because `<body>` does not exist yet at that point. Only a colour is let through: the value is written by our own code from what Python sends, and a stylesheet is not the place to find out otherwise.

For a first load in a fresh session there is nothing remembered, so the CSS carries a `prefers-color-scheme` fallback. A guess, but the right one for the common case of a dark app on a dark desktop, and better than white either way.

## Measured

Sampling every frame, both toggles, dark theme:

| | white frames |
|---|---|
| before | 3 per toggle |
| after | **0** across 1141 frames |

Path finder behaves the same, and light mode stays white throughout (1082 frames, one distinct colour).

## A wrong turn worth recording

My first attempt wrapped the display-options band and the path panel in always-present containers, on the theory that the toggle moved the component in Streamlit's element tree and Streamlit rebuilt the iframe (issue #189). A control run on unmodified `main` showed the iframe surviving the toggle there too - so the theory was wrong and the change bought nothing measurable. It is not in this PR; the 300-line re-indent it needed would have been pure cost.

`pytest` 1874 passed, `ruff` clean. No Python changed - this is the vendored viewer only, which is hash-versioned, so the new document serves under a fresh URL rather than from a stale cache.

### DIFF
--- a/orionbelt_ontology_builder/lib/graph_viewer/index.html
+++ b/orionbelt_ontology_builder/lib/graph_viewer/index.html
@@ -2,8 +2,41 @@
 <html>
 <head>
 <meta charset="utf-8">
+<script>
+// Paint the background before anything else is parsed.
+//
+// This document is replaced whenever Streamlit re-mounts the component — which
+// a click on the Display options or Path finder switch is enough to do — and a
+// fresh one used to paint the stylesheet's white until the theme arrived over
+// the round trip from Python. Two or three frames of white, in the middle of a
+// dark app, on a control people toggle constantly (issue #372).
+//
+// The colour is the one this component last drew, kept per session and per
+// origin. The stylesheet below carries the fallbacks for a first-ever load,
+// where nothing is remembered yet.
+(function () {
+    try {
+        var bg = sessionStorage.getItem('viz_theme_bg');
+        // Straight into a stylesheet, so it applies to a <body> that does not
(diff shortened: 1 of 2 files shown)


SuperCowProducts:
All good after `git merge origin/fix/graph-iframe-white-flash-372`!

Issue #381, opened 2026-09-01
[UX] make more room for visualisation graph


SuperCowProducts:
Changing the height was eventually abolished at some commit, and this does actually simplify movement in OrionBelt, although the space is a lot more cramped now:

<img width="2556" height="1600" alt="Image" src="https://github.com/user-attachments/assets/c09df264-48af-45ba-8ccf-af594e68b006" />

I'm not quite sure what the best way to improve on this would be at this stage in development, but here are 3 things I was thinking:
1. Bring back height adjustment (although reaching out for options will require scrolling again, and in general working in visualisation feels less 'stable' in a way)
2. Provide all the Interactive Graph menus (Display options, Path finder, etc.) in full screen (perhaps in a way so that they can be added or removed from view, had thought of that in https://github.com/ralforion/orionbelt-ontology-builder/issues/177, but unsure how practical that is, although would definitely be very cool): I am usually exiting fullscreen a lot only to access the menus.
3. Reduce empty space above (although I think this may be difficult without interfering with the layout on the rest of the sections)

ralfbecher:
Landed in #382 (merged as 7e864cb).

**Menus in fullscreen (idea 2).** Fullscreen used to be taken on the canvas' own container, which lives inside the component iframe, so every control stayed outside it. It is now taken on the page, with a class on the body that hides the sidebar, Streamlit's header, the breadcrumb, the page title and the section tabs. Display options, Find & focus, Path finder, Node options, the details panel and the status bar all come along, so there is no more exiting fullscreen just to reach a menu. Selections round-trip while fullscreen too, so the details panel behind the graph is the live one.

**More room (idea 3).** Three things, none of which touch the page's own spacing:

* A third band switch, Find & focus, for the row that finds an entity, filters what is drawn and focuses on one node. Off, it hands its 66px to the canvas and remembers the setting like the other two switches.
* Render sized to its label instead of stretched across its column, taking the switch row from 40px to 28px.
* A real bug: the rule that keeps empty CSS-only blocks out of the layout had stopped matching after a Streamlit DOM change. Measured in the browser, 0 elements matched while 4 sat above the graph, each holding a 16px slot.

Measured at 1440x900: the canvas goes from 540px to 606px with the bands off, and to 696px in fullscreen with every control still on screen.

[...]

PR #382, opened 2026-09-01
feat(viz): fullscreen keeps the graph's controls, and the page gives it more room (#381)

ralfbecher:
Addresses two of the three ideas in #381: the menus in fullscreen, and more room for the canvas. Manual height (idea 1) is deliberately not brought back, since the canvas already takes everything below the controls and a slider could only give it less.

## Fullscreen keeps the controls

Fullscreen was taken on the canvas' own container, inside the component iframe, so every control stayed outside it. It is now taken on the page, with a class on the body that hides the sidebar, Streamlit's header, the breadcrumb, the title and the section tabs. Display options, Find & focus, Path finder, Node options, the details panel and the status bar all come along.

This also retires the workaround from #189: the fullscreen element used to be destroyed by any rerun that re-mounted the component, so selections had to be held back until exit. The page survives every rerun, so a node click in fullscreen updates the panel that is now on screen.

## Room above the canvas

- A third band switch, **Find & focus**, for the Find / Node options / Focus row. Off, it hands its 66px to the graph.
- **Render** sized to its label rather than stretched across its column: the switch row drops from 40px to 28px.
- The rule that takes empty CSS-only blocks out of the layout had stopped matching after a Streamlit DOM change. Measured in the browser: 0 elements matched while 4 sat above the graph, each holding a 16px slot.

The page's own spacing (title, tabs, row gaps) is untouched.

## Also

The canvas toolbar's tooltips never appeared: they were native `title` attributes and the hover target is the `<svg>` inside the button. The page draws them itself now.

## Verified in the running app (1440x900)

- Canvas 540px, 606px with the Find row off, 752px in fullscreen with the controls still on screen.
- Fullscreen survives a panel toggle and a node click; the Esc path puts the chrome back and resizes the canvas.
- Sidebar, header, breadcrumb, title and tabs all hidden in fullscreen; every control visible.

1888 tests pass; ruff and mypy clean.

Note for review: the fullscreen box was measured in a headless browser, where the window is already the screen. Worth one look in a real window smaller than the screen.

### DIFF
--- a/orionbelt_ontology_builder/lib/graph_viewer/index.html
+++ b/orionbelt_ontology_builder/lib/graph_viewer/index.html
@@ -773,6 +773,24 @@ function setNativeFsExitHook(fn) {
     try { if (window.top) window.top[TOP_FS_EXIT_HOOK] = fn; } catch (e) {}
 }
 
+// The hook is a function belonging to *this* document, and fullscreen now
+// outlives it: Streamlit re-mounts the component and the closure the old
+// document installed goes with it, so the desktop window leaving fullscreen
+// would have nobody left to tell and the chrome would stay hidden. Every mount
+// that finds the page still fullscreen claims it again (review P2).
+function installNativeFsExitHook() {
+    setNativeFsExitHook(function () { if (stageOn()) leaveFullscreen(true); });
+}
+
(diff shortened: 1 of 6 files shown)


SuperCowProducts:
Very nice!

Issue #389, opened 2026-09-01
[suggestion] deactivate 🎯 notifications when changing focus


SuperCowProducts:
<details>

<summary>
original
</summary>

I think these notifications look really cool, but I actually find them quite distracting whilst working in the ontology. If I figure out a good way to do remove them by modifying the code I'll post a patch here.

</details>

[UPDATE]

There, this does the trick:

<details>

<summary>
patch
</summary>

```patch
diff --git a/orionbelt_ontology_builder/views/visualization.py b/orionbelt_ontology_builder/views/visualization.py
index baaddc1..2035ad6 100644
--- a/orionbelt_ontology_builder/views/visualization.py
+++ b/orionbelt_ontology_builder/views/visualization.py
@@ -2786,21 +2786,21 @@ def render_visualization():
             st.rerun()
         # Past that rerun, so it survives to reach the user: what the filter
         # kept out of the view when it was created (issue #194).
-        for _announcement in (
-            st.session_state.pop("_viz_new_hidden_announce", None) or []
-        ):
-            st.toast(_announcement, icon="🙈")
+        # for _announcement in (
+        #     st.session_state.pop("_viz_new_hidden_announce", None) or []
+        # ):
+        #     st.toast(_announcement, icon="🙈")
         # Focus mode ended itself because nothing was left to focus on — the
         # entity went, its type was switched off, or there was nothing to derive
         # a first seed from. Silently un-ticking the box is the confusing half of
         # not standing an arbitrary entity in (issue #335), so say so, and say
         # what starts a focus instead.
-        if st.session_state.pop("_viz_focus_left_note", False):
-            st.toast(
-                "Focus off: nothing left to focus on. Ctrl/Cmd-click a node to "
-                "start a new one.",
-                icon="🎯",
-            )
+        # if st.session_state.pop("_viz_focus_left_note", False):
+        #     st.toast(
+        #         "Focus off: nothing left to focus on. Ctrl/Cmd-click a node to "
+        #         "start a new one.",
+        #         icon="🎯",
+        #     )
         # Say what is being held back, small, right above the canvas: a focus or
         # a narrowed filter is otherwise invisible, and an entity that isn't
         # drawn looks lost rather than filtered (issue #222 follow-up).
@@ -2971,7 +2971,7 @@ def render_visualization():
                             _note = f"Focusing on {_focus_label}" + (
                                 " only" if _replace else ""
                             )
-                        st.toast(_note, icon="🎯")
[...]

Issue #335, opened 2026-08-28
[suggestion] allow `Focus node(s)` to be completely empty


SuperCowProducts:
Currently displays the first class in alphabetical order, probably as a 'protective' measure, but actually it's very nice to be able to just clear everything and start over: this already happens outside of `Focus on one node` when `Clear all` is pressed. After all, there's no harm in not displaying anything: in fact, 99% of the time you don't need the first element in alphabetical order to show by default!

ralfbecher:
Shipping this as #336, though not quite as asked, so worth explaining.

An empty `Focus node(s)` is not allowed. Instead, clearing it ends focus mode. The reason is that focus mode with no seeds does not draw nothing today, it draws everything:

```python
focus_pruning = bool(focus_mode and focus_seed_ids)
```

and `graph_node_cap` spells out why: focus mode is allowed to assemble more nodes than can be drawn *only* because the prune cuts it back afterwards. With no seeds nothing prunes. So the backfill you spotted was not really a protective measure around the picker, it was what kept that state unreachable and the node cap (#216) honest. Deleting it on its own would have made Clear all dump the whole ontology onto the canvas, past the draw cap.

Ending the mode gets you the same thing: clear it, everything comes back, and Ctrl/Cmd-clicking a node starts a fresh focus. It is also where the last Ctrl/Cmd-click already lands after #328, so both ways out of a focus now agree.

Your report turned out to cover more than the picker. The same backfill fired whenever the saved seeds resolved to nothing on a render, so focusing a data property and then unticking `Data Props` moved the focus onto the first class instead. That was the quieter half, and it is gone too: `focus_labels[0]` no longer exists anywhere.

One thing that fell out of removing it: with nothing to derive a first seed from, ticking `Focus on one node` would just un-tick itself, which looks like a broken control. It now says `Focus off: nothing left to focus on. Ctrl/Cmd-click a node to start a new one.`

If you genuinely want an empty focus with a blank canvas rather than the mode ending, say so and we can look at it, but it means teaching the build that an empty focus draws nothing, which is a larger change than it looks.

SuperCowProducts:
Sounds like an intelligent design choice - further use of OrionBelt with the new features will confirm!

PR #393, opened 2026-09-02
fix(viz): a focus click is applied without a toast about it (#389)

ralfbecher:
Addresses #389.

## What was wrong

Ctrl/Cmd-clicking a node in the graph raised a toast in the top right corner saying what had just been clicked: `Focusing on Class: Person`, `Dropped Class: Person from the focus`, `Focus off`. It is a second telling of something the user did on purpose, it sits there for four seconds at a time (measured: added 157ms after the click, removed at 4160ms), and it repeats all through a working session, which is what #389 is about.

### It was also wrong as soon as you worked at speed

A toast is reused for the same point in the script, and one already on screen is not re-shown. So a second focus click inside those four seconds changed nothing in the corner. Measured on `main`, cmd-clicking real nodes on the canvas:

| t | event | toast in the corner |
| --- | --- | --- |
| 0.03s | cmd-click `Person` | `🎯 Focusing on Class: Person` |
| 1.05s | cmd-click `Department`, toast still up | still `...Person` |
| 1.3s to 3.8s | | still `...Person`: one distinct toast text across the whole window |
| 4.09s | | gone, and no toast for the second click ever appeared |

The second click had applied all along: the focus ended at `Focused on Person · 1 hop`, meaning `Department` had been taken back out of it while the corner still read "Focusing on Class: Person". So for the second click the toast said nothing, and what it did say named the wrong node. The note has no such window; it was correct at every sample.

## What changed

The click is applied and the page reruns, with nothing said about it. What was already there says more, and stays:

- The canvas redraws to the focused neighbourhood.
- The note on the Node options label reads `Focused on Person, Department · 1 hop · 12 hidden by focus`. It names the seeds (up to five, then a count), the depth and what is held back, it is visible whether the panel is open or closed, it does not fade, and it keeps up with clicks that come faster than a toast can.

The other toasts on the page are left alone, and each for its own reason:

- **Focus ended by itself** (`🎯 Focus off: nothing left to focus on`) is not something the user did. Un-ticking the box in silence is precisely what issue #335 was filed about.
- **What a filter kept out of the view** (`🙈`) is the same kind of thing: state the user did not set, and it is announced as it happens (issue #194).
- **A modifier-click that did nothing** (`ℹ️ Focus is available for classes, individuals, and SKOS concepts`) explains why nothing happened, which nothing else on the page does.

## Verified

Live in the running app, on the org template:

- Modifier-click on `Person`: no toast, the note reads `Focused on Person · 1 hop`, and the graph redraws to the focus.
- Modifier-click on `Person` again, which takes it back out: no toast, and the note goes empty, focus off.
- The same measurement run against `main` for a control, since an absent toast proves nothing unless the toast was there to begin with: a DOM observer catches it being added 157ms after the click and removed at 4160ms, at full opacity throughout. On this branch the observer sees nothing.

`pytest` (1913 passed), `ruff format --check`, `ruff check` and `mypy` are all clean. `tests/test_viz_focus_click.py` gains two source-level pins: one that this branch applies the click and says nothing, one that the focus-that-ended-itself toast is still there. They are pinned at the source because the click arrives through the graph component, which does not render under AppTest.

## Note on scope

The patch in #389 also comments out the other two toasts. Those are the ones that report state the user did not cause, so they are kept here. If they are unwanted as well, the better answer than deleting them is a persisted "Graph notifications" setting alongside the other display settings, and that can be a separate change.

### DIFF
--- a/orionbelt_ontology_builder/ui.py
+++ b/orionbelt_ontology_builder/ui.py
@@ -1636,10 +1636,10 @@ def focus_seeds_after_request(seeds, label, replace=False, focus_on=False):
 def viz_apply_focus_click(label, replace=False):
     """Apply a modifier-click on ``label`` to the focus state in session.
 
-    Returns the ``(seeds, focus_on)`` it wrote, so the caller can word the
-    toast. Split out from the page so the state it touches can be tested: the
-    click arrives through the graph component, which does not render under
-    AppTest, so nothing else can reach this path.
+    Returns the ``(seeds, focus_on)`` it wrote. Split out from the page so the
+    state it touches can be tested: the click arrives through the graph
+    component, which does not render under AppTest, so nothing else can reach
+    this path.
 
     Mode and seeds are written together, never across renders: focus mode with
     no seeds backfills an arbitrary first label, so a pass through that state
--- a/orionbelt_ontology_builder/views/visualization.py
+++ b/orionbelt_ontology_builder/views/visualization.py
@@ -2980,20 +2980,16 @@ def render_visualization():
                     _id_to_label = {v: k for k, v in focus_targets.items()}
                     _focus_label = _id_to_label.get(selection.get("nodeId"))
                     if _focus_label:
-                        _replace = bool(selection.get("replace"))
-                        _seeds, _focus_on = viz_apply_focus_click(
-                            _focus_label, replace=_replace
+                        # Applied without a word about it. A toast confirming a
+                        # click the user just made is a second telling: the
+                        # canvas redraws to the focus, and the standing note on
+                        # the Node options label names the seeds, the depth and
+                        # what is held back — and stays there instead of fading
+                        # (issues #389, #222 follow-up). What is still said here
+                        # is the case below, where the click did nothing.
+                        viz_apply_focus_click(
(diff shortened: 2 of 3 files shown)


SuperCowProducts:
@ralfbecher if you do want to implement a "persisted "Graph notifications" setting" I'll definitely use that, otherwise a tweak in the source code similar to what I did in the issue is straightforward for anyone who wishes not to see these toast notifications for focus clicks.

SuperCowProducts:
In fact, in the updated source, getting rid of the focus click notifications is as simple as commenting out one line in a file:

<details>

<summary>
patch
</summary>

```patch
diff --git a/orionbelt_ontology_builder/views/visualization.py b/orionbelt_ontology_builder/views/visualization.py
index 5cd36bf..0eed2f0 100644
--- a/orionbelt_ontology_builder/views/visualization.py
+++ b/orionbelt_ontology_builder/views/visualization.py
@@ -2993,7 +2993,7 @@ def render_visualization():
                             _note = f"Focusing on {_focus_label}" + (
                                 " only" if _replace else ""
                             )
-                        st.toast(_note, icon="🎯")
+                        # st.toast(_note, icon="🎯")
                         st.rerun()
                     else:
                         st.toast(
```

</details>

PR #421, opened 2026-09-06
fix(ui): an edit confirmation clears in two seconds, not four

### LINKED ISSUE #419: [minor] Relation updated! stays on screen too long
I think 2 secs is more than enough. But should be a straightforward patch if user wants to.

ralfbecher:
Closes #419.

## The report

"Relation updated!" stays on screen too long. Two seconds is more than enough.

## The change

`display_flash_message()` passes `duration=2` to `st.toast` for the success confirmations. Streamlit's default is `"short"`, which is four seconds.

These carry one piece of news each: the edit landed. That is read at a glance, and what is left of the four seconds is a message about the last edit sitting over the next one. All six toast call sites that go through the flash are of that kind (`Relation updated!`, `Restriction updated!`, `Annotation updated!`, `Annotation deleted!`, and the two entity deletions).

A warning or an error given as a toast keeps the default, since neither is read that fast. Nothing today sends one, so this is a guard on the next one rather than a behaviour change.

`duration` is a Streamlit 1.62 parameter, which the pin is already on.

## Tests

`tests/test_flash_toast.py` gains two cases: a confirmation asks for two seconds, and a warning keeps the default. The fixture's `st.toast` stand-in now records the duration.

Full suite: 2031 passed. `ruff` and `mypy` clean.

### DIFF
--- a/orionbelt_ontology_builder/ui.py
+++ b/orionbelt_ontology_builder/ui.py
@@ -3020,6 +3020,13 @@ def show_message(message: str, type: str = "info"):
 #: Icons for the toast form of a flash message, by message type.
 _FLASH_TOAST_ICON = {"success": "✅", "warning": "⚠️", "error": "🚫", "info": "ℹ️"}
 
+#: How long a confirmation toast stays up, in seconds. Streamlit's default is
+#: four, and these carry one piece of news each: the edit landed. That is read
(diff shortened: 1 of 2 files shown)


SuperCowProducts:
Thanks!

Issue #437, opened 2026-09-14
[performance] oxigraph + cytoscape.js


SuperCowProducts:
Obviously this would represent a major undertaking, although it was a long-term transition I was thinking of after experiencing a bit of lag when editing an ontology with 1000+ entities. In fact, it sounds like this would basically be a rewrite at this stage. Note the choice of the 2 libraries was heavily influenced by a conversation with Claude, not with real-world metrics: if I ever end up trying this out though, it may bring about something useful. Apparently oxigraph is rdflib-compatible, and the idea of cytoscape was to bypass any slowdowns that may be caused by streamlit bottlenecks. Out of all options building from OrionBelt felt like the path that made most sense, and it might be merged back into OrionBelt, since the idea is to improve performance and scalability, not remove OrionBelt's brilliant GUI. Just thoughts for now, nothing concrete.

ralfbecher:
Thanks for writing this up. I had a look at where the app actually spends its time on a large ontology before thinking about libraries, and I believe the lag is not in rdflib or vis-network as such. Three things in the current code scale with ontology size on every click:

- **Undo snapshots.** Every edit takes a checkpoint (there are about 80 call sites), and each checkpoint serializes the whole graph to N-Triples in pure Python. Undo re-parses it from scratch. For a 1000-entity ontology that is tens of thousands of triples per click. See `take_snapshot` in `ontology_manager.py`.
- **Uncached per-rerun queries.** Streamlit reruns the page script on every widget interaction. The sidebar recomputes the statistics each time, and the list pages call `get_classes` and friends each time, with several index lookups per entity. None of it is memoized against a graph revision.
- **The Source page** serializes the entire ontology to Turtle and renders it as a line-numbered code block on every rerun.

On the two libraries:

**Oxigraph** is rdflib-compatible through the `oxrdflib` store plugin, and adopting it is close to a one-line change (`Graph(store="Oxigraph")`) rather than a rewrite. But the store only accelerates triple pattern matching and SPARQL. Serialization and parsing still go through rdflib's Python layer, and every matched triple still crosses the Rust/Python boundary as an rdflib object. So it would help the SPARQL page, help the list pages a bit, and leave the snapshot cost untouched. It would also add a native extension the desktop build has to bundle.

[...]

PR #449, opened 2026-09-16
fix(ui): the page shim lets go of the parent document when its frame goes

ralfbecher:
From the follow-up on #437: memory grows over a session. Part of that was the stray server processes (#448); this is the part inside a running instance, and it is in the browser, not the server.

## Server side

The Python side does not grow. Driving the app through AppTest on the reporter's ontology (1,098 classes, 6,323 triples) for 140 reruns across the Visualization page unchanged, a rename with an undo checkpoint each time, and Classes/Visualization page switches: Python-tracked memory stayed at 59 to 61 MB throughout, session state kept its 81 keys, and the largest entry stayed at 0.3 MB. Process RSS moved between 330 and 395 MB and plateaued, which is allocator churn over the per-rerun payloads, not retention. The revision-keyed memo cache clears on every revision change and the undo history is capped at fifty diffs, so nothing there accumulates.

## Browser side

Measured in Chromium with a forced GC before each reading, toggling a graph setting to rerun the Visualization page:

| | per rerun, before | per rerun, after |
|---|---|---|
| retained DOM nodes | about 240 to 280 | 0 (2,061 to 2,070 over 30 reruns) |
| live iframe elements | +1 | about +1 per 10 reruns |
| event listeners | +1 | 0 |

A stock Streamlit app with the same widgets, tooltips and coloured markdown is flat on all three, so this was ours. A heap snapshot showed the retained nodes were the markdown and metric elements Streamlit had replaced on earlier reruns, plus one detached component iframe per rerun.

## Cause

`_HELP_WIRING_JS` (the help text and tab order wiring in `page_shim_html`) runs inside a component frame and registers a `MutationObserver` on the parent document's body. It was never disconnected, on the assumption that the frame is mounted once. The frame is re-created whenever its arguments change, and its arguments are the page's help texts, so on a busy page that is most reruns. An observer left behind by a frame that is gone never delivers again, but it stays registered: it keeps that frame's whole document alive, and it queues a record for every node Streamlit removes on every later rerun, which nothing ever drains. That is the one iframe and the few hundred nodes per rerun.

## Change

The new frame disconnects the previous observer (kept on the parent document, the same pattern the Enter shim already uses for its handler), and each frame disconnects its own observer on `pagehide`. The Enter shim also removes its handler on `pagehide`, so a frame that is never replaced leaves nothing behind either.

The residual iframe wrapper every ten or so reruns is an empty JS wrapper with no document behind it (the node count is flat), and is not chased here.

## Tests

`tests/test_help_tab_order.py` pins the replacement, the `pagehide` release and the Enter shim's release. Full suite 2,097 passed, ruff and mypy clean.

Refs #437.


ralfbecher:
Review finding addressed: `pagehide` also fires on the way into the back/forward cache, and a restored page does not rerun its frames' scripts, so releasing there left the restored page without the help wiring and the Enter override.

Rather than reinstalling on `pageshow`, both handlers now stand down when `event.persisted` is true. A cached page comes back with its whole tree, the frame and its registrations included, so there is nothing to put back; the release is only needed for a frame that is really going away. `test_a_page_cached_for_back_forward_keeps_both_shims` pins it for both shims.

Issue #141, opened 2026-07-17
[idea] update only necessary part of graph after `Ctrl`+`Click`


SuperCowProducts:
I often want to gradually 'roll out' my ontology in a specific areas, which means hopping from node to node with `Ctrl`+`Click` - because the graph needs to update though, it keeps on jumping around the screen just like I jump around from node to node, so if there were a way to have it stay put (save for the new items that need to sprout up) I think that would really improve the experience.

PR #464, opened 2026-09-20
fix(viz): a re-layout keeps the view you are working in

### LINKED ISSUE #460: [suggestion] don't zoom out when re-rendering graph
I can find my place in the graph again if I had selected the item as it stays highlighted, which is very handy, just that having to zoom in every time is a bit annoying. I think it's better to leave the decision to the user, or even a separate 'zoom out to whole ontology button' is fine.

ralfbecher:
Closes #460.

## What was wrong

Render, and a node-spacing change with it, ended on a fit: the graph was laid out again and the camera zoomed out to the whole ontology. As the reporter puts it, the selection surviving is handy, because it is how you find your place again, but zooming back in every time is not.

Measured on `main`, FOAF, zoomed to 1.345:

| action | camera |
| --- | --- |
| Render | 1.345 to 0.835 |
| Node Spacing | 1.345 to 0.837 |
| a filter toggle, an add, a delete | held (the #455 fix) |

So this is the one path that was left.

## What changed

The layout is deterministic (the random seed is pinned, #141), so a Render on the same nodes puts them back exactly where they were. The camera from the last settled render now carries across instead of being thrown away. A first visit has no camera to carry and still fits, which is what puts the graph on screen at all.

A fresh layout does place every node again, so a wide spacing change can move the graph out from under the camera. That is the question the node-set rebuild already asks (#455), so it is asked here too, once physics has settled and the real positions are known: if the view ends up pointing at nothing, fit after all. A camera the user moved themselves during the pass is left alone.

## The Fit button

Zooming out to the whole ontology was only ever a side effect of Render, so taking it away would leave no way to do it on purpose. It gets a control of its own, which is also the reporter's own suggestion: a Fit button on the canvas beside Maximize, Download and Copy. It animates, and it teaches the viewport cache the new frame so a re-mount keeps it.

The icon is the four-arrows "zoom out to fit" glyph. It sits next to Maximize, whose corner-bracket icon is a cousin; the tooltips tell them apart, but say the word if you would rather have a different glyph or a different order in the strip.

## Verification

Driven in the running app on FOAF:

- Render at zoom 1.974 leaves scale and position bit for bit unchanged (it landed at 0.835 before)
- a Node Spacing step likewise: 1.216 in, 1.216 out
- camera parked on empty canvas at (9000, 9000), then Render: fits back, all 26 nodes in frame
- the Fit button: 1.97 to 0.83, and the cached viewport matches what is on screen, so toggling the details panel keeps it

Not touched: vis rescales the view when the canvas height changes (the details-panel toggle moves it between 0.774 and 0.825 and back). That is vis keeping the same visible extent across a resize, it predates this change, and the cache is untouched by it.

## Tests

`tests/test_viz_camera_hold.py` gains the re-layout hold, the first-visit fit and the rescue. The old `test_a_fresh_layout_still_frames_the_graph` said the fit applies to "a Render click or a first visit", which is no longer true, so it is now `test_a_first_visit_still_frames_the_graph` and asserts the exact condition.

`tests/test_viz_fullscreen.py` covers the new tooltip and gains a guard that every button carrying a `data-tip` is wired to the tip placement, which fails if the new button is left out. `tests/test_viz_overlays.py` includes it in the toolbar-clearance geometry.

Full suite (2113), ruff and mypy green.


ralfbecher:
Fixed in cef9b0e.

The finding is right, and the mechanism is the one the cache's own comment already records: sessionStorage survives a reload while `viz_render_seq` restarts at 0, so the first render after a reload lands in the fresh-layout branch holding the previous session's frame. Both sibling branches check the cache's identity before trusting it; that branch checked nothing.

The gate is the node hash, which the cache is already written under:

```js
savedView = (_raw && _raw.hash === nodeHash && _raw.view) || null;
```

A matching hash says the saved view was taken of the same nodes, and the same nodes laid out again from the same pinned seed land in the same places, so the frame still means what it meant. That keeps an explicit Render and a spacing change on the same graph holding the view, and sends a different graph back to being framed from scratch.

I did not add a separate session or generation guard on top. The generation counter is what a reload resets, so it cannot be the thing that distinguishes "same graph" across one, while the hash can and does. If a reload should fit rather than land you where you were even on an identical graph, that is a preference worth its own issue.

## Reproduced both ways

A cache carrying a foreign hash and a 6x view centred on the origin, so nodes do fall inside it and the rescue fit deliberately stays quiet, which is the accidental-overlap case in the report:

| | before the gate | after |
| --- | --- | --- |
| Render with that stale cache | inherits 6x at (0, 0) | fits at 0.78 |
| Render on the same graph | holds 1.329 | holds 1.329 |

`tests/test_viz_camera_hold.py` gains `test_a_carried_camera_has_to_be_one_of_this_graph`, which also pins that the gate reads the same key the cache is written under. Full suite 2114, ruff and mypy green.

